Issue #15, opened 2020-03-20
Configuration


srid:
Add `config.dhall` in the notes directory, for specifying things like site title, URL, theme color.

Configuration should be optional, with sensible defaults.

- [ ] z-index.html should be `index.html` by default unless configured to point to elsewhere
  - Verify that `${zindexname}.md` does not exist before writing its HTML
- [ ] Make mathjax include configurable

PR #16, opened 2020-03-20
Math support

0x83f:
Support for rendering math would be nice, I hope the MathJax dependency is not a problem

### DIFF
--- a/src-bin/Main.hs
+++ b/src-bin/Main.hs
@@ -46,6 +46,7 @@ renderPage site r val = html_ [lang_ "en"] $ do
     stylesheet "https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.11.2/css/all.min.css"
     style_ [type_ "text/css"] $ C.render style
     googleFonts $ [headerFont, bodyFont, monoFont]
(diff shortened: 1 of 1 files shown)


src/Neuron/Zettelkasten/View.hs
```diff
@@ -91,7 +92,10 @@ renderZettel (store, graph) zid = do
   div_ [class_ "zettel-view"] $ do
     div_ [class_ "ui raised segment"] $ do
       h1_ [class_ "header"] $ toHtml zettelTitle
-      MMark.render $ useExtension (linkActionExt store) zettelContent
+      MMark.render
+        $ useExtension (linkActionExt store)
+        $ useExtension (mathJax (Just '$'))
+        $ zettelContent
```
srid: Hey, thanks for the PR! Since neuron uses `rib`, and as `rib` already includes mathJax in [its list of default extensions](https://github.com/srid/rib/blob/48c0e70/src/Rib/Parser/MMark.hs#L119), this part of the change is not necessary. Can you confirm that it works without it (and the cabal file change)? 

If yes, then we only need to merge the `script_` include of mathjax Javascript. I'm not against including it; but I might disable its inclusion _by default_ unless explicitly configured by the user (see #15).

0x83f: You are right, the script include is sufficient. An option to enable it explicitly would be nice though

Issue #18, opened 2020-03-21
Cluster detection


srid:
As the Zettelkasten grows, notes will naturally belong to specific clusters. For eg., science-related notes would be part of one cluster, whereas personal journal-related notes would be in an entirely different cluster. 

In computer science, we call this a [connected component](https://en.wikipedia.org/wiki/Component_(graph_theory)) of a graph. Pictured below is a graph with three connected components.

Neuron's z-index should distinctly display these components. If there are three components, for example, there should be three separate trees/forests.

![](https://upload.wikimedia.org/wikipedia/commons/thumb/8/85/Pseudoforest.svg/480px-Pseudoforest.svg.png)

gillchristian:
I was wondering how to add multiple "entry points" to the `z-index`, the the moment only `index.md` is considered one.

Would this _solve_ that?

srid:
At the moment, my multiple entry points are all linked from the top-level entry point `index.md`. 

With this feature, however, I hope to obviate `index.md` and have these entry points naturally emerge from the zettelkasten connections (i.e., without _explicitly_ designing them as an "entry point").

I am yet to fully flesh this feature out, but I imagine each cluster to have an (optional) single entry point -- i.e., if there is exactly one [mother vertex](https://www.geeksforgeeks.org/find-a-mother-vertex-in-a-graph/), it can automatically be promoted to be its entry point.

Issue #5, opened 2020-03-08
"Edit this Zettel on Github"


srid:
Add a link that will take users to the zettel on Github, allowing them to edit it.

Requires #15 (for `editUrl` configuration)

PR #23, opened 2020-03-23
feat(nav): Add z-index on the 'Navigate up' section

gillchristian:
This PR adds a link to the `z-index` in the "Navigate up" section as a quick way to go back to the graph view.

Screenshot 2020-03-23 at 19 04 38

NOTE: I'm not familiar with Rib and the Nueron codebase. Please let me know what should be changed.

### DIFF
--- a/src/Neuron/Zettelkasten/View.hs
+++ b/src/Neuron/Zettelkasten/View.hs
@@ -26,7 +26,7 @@ import Neuron.Zettelkasten.Graph
 import Neuron.Zettelkasten.ID
 import Neuron.Zettelkasten.Link (linkActionExt)
 import Neuron.Zettelkasten.Link.Action (LinkTheme (..))
-import Neuron.Zettelkasten.Link.View (renderZettelLink)
+import Neuron.Zettelkasten.Link.View (renderZettelLink, renderLink)
 import Neuron.Zettelkasten.Route
 import Neuron.Zettelkasten.Store
 import Neuron.Zettelkasten.Type
@@ -87,6 +87,9 @@ renderIndex (store, graph) = do
     h2_ "Tree"
     ul_ $ renderForest Nothing LinkTheme_Default store graph forest
 
+zIndexLabel :: Text
+zIndexLabel = "z-index"
+
 renderZettel :: Monad m => (ZettelStore, ZettelGraph) -> ZettelID -> HtmlT m ()
 renderZettel (store, graph) zid = do
   let Zettel {..} = lookupStore zid store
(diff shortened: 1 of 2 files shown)


humaidq:
It is nice to have a link to the z-index somewhere, but having it in
the "Navigate up" section doesn't feel right to me. Maybe it would be
make more sense for it to be in the footer?

gillchristian:
@humaidq I didn't find a place to put it that wouldn't require bigger layout/markup changes. IMHO the graph is the top level so I find it intuitive to be part of 'Navigate up'.

That said, I'm fine with it being anywhere in the footer.

There's an open issue for 'Edit on GitHub' (#5), maybe the footer will be extended with more things and a better place from UX point of view.

srid:
I like the idea of including "z-index" in a place like this. I agree that it feels odd to place it in "Navigate up" section (though that would *eventually* make sense once "z-index" becomes part of the graph itself, as a result of #18).

For now, I'd introduce a separate footer section below both Connections/NavigateUp (still inside the same teal box perhaps) and put the z-index in there.

src/Neuron/Zettelkasten/View.hs
```diff
@@ -105,7 +105,11 @@ renderZettel (store, graph) zid = do
         div_ [class_ "column"] $ do
           div_ [class_ "ui header"] "Navigate up"
           let forestB = obviateRootUnlessForest zid $ dfsForestBackwards zid graph
-          ul_ $ renderForest Nothing (LinkTheme_Simple $ Just zid) store graph forestB
+          ul_ $ do
+            li_ $ do
+              a_ [class_ "zettel-link item", href_ "/z-index.html", title_ "z-index"] $ do
+                span_ [class_ "zettel-link-title"] "z-index"
```
srid: Respecting the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), I'd extract [this code](https://github.com/srid/neuron/blob/44365c6/src/Neuron/Zettelkasten/Link/View.hs#L61-L66) into a function, and then re-use it here.

gillchristian: Extracted the two links, active and regular. Although active is not a link is used in place of one.

They aren't exactly the same to justify using the same function I think.

gillchristian:
Went for "Others" as the section's title, but I'm not quite happy about it. Wdyt?

Screenshot 2020-03-23 at 23 01 42

srid:
@gillchristian I've made some tweaks to your branch at https://github.com/gillchristian/neuron/pull/1 ... if you could merge that PR, I'll merge this one right away.

srid:
Actually, I've merged it from the command line. Thanks for the PR!

Issue #42, opened 2020-03-29
Enable queries from the CLI


felko:
Add a `neuron ... query` command to the CLI, which would ease editor support. It would return a list of Zettel IDs that match the specified query. Example: `neuron ... query --tag=science`

Maybe use this to find zettels by title or content with ripgrep, sort of like `neuron ... search` does but returning only the list of IDs instead of starting a fzf interface.

srid:
Closed by https://github.com/srid/neuron/pull/51

srid:
Documented: https://neuron.srid.ca/2013501.html

PR #47, opened 2020-04-02
Implement queries from the CLI

felko:
Implement `neuron ... query` command as described in #42 

About this suggestion:
> Maybe use this to find zettels by title or content with ripgrep, sort of like neuron ... search does but returning only the list of IDs instead of starting a fzf interface.

It would be easier to implement it as a `ByTitle` constructor in the `Query` type so it's not implemented in this PR.


srid:
Already merged at https://github.com/srid/neuron/commit/e0d9e7bbc0d2a2cf841d60f7fe5fd36f9f425bca

Issue #43, opened 2020-04-01
Add `-e` option to automatically open in editor


srid:
The following:

```
neuron ./notes new "Some title" -e 
neuron ./notes search -e
```

Should automatically open the file using `$EDITOR`.

Should be equivalent to piping to `xargs -rt $EDITOR`.

**Tasks**

- [x] Add `-e` to the `new` command
- [x] Add `-e` to the `search` command

PR #52, opened 2020-04-03
Add `-e` / `--editor` flag for `neuron … new` command

Gabriel439:
Related to https://github.com/srid/neuron/issues/43

### DIFF
--- a/src/Neuron/Zettelkasten.hs
+++ b/src/Neuron/Zettelkasten.hs
@@ -1,6 +1,8 @@
+{-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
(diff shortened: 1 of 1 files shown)


src/Neuron/Zettelkasten.hs
```diff
@@ -150,14 +159,32 @@ generateSite writeHtmlRoute' zettelsPat = do
...
+
+      if edit
+        then do
+          maybeEditor <- Env.getEnv "EDITOR"
+
+          editor <- case maybeEditor of
+            Nothing -> do
+              putStrLn "Set the EDITOR environment variable"
+
+              Exit.exitFailure
```
gillchristian, line 183: [Q] maybe the path of the new fine should still be printedv

srid: I agree; it should print the new file path regardless. If the user gets this error, and they don't see the path to the created filepath, they wouldn't easily know how to open it manually.

Issue #40, opened 2020-03-29
Add --version CLI argument


srid:
It must also show the git revision.

PR #55, opened 2020-04-04
Add --version CLI argument

TristanCacqueray:
Fixes: https://github.com/srid/neuron/issues/40

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -25,6 +25,7 @@ let
     GITDESC=`cat ${gitDescribe}/output`
     cat << EOF > $out/src/Neuron/Zettelkasten/Version.hs
     module Neuron.Zettelkasten.Version where
+    version :: String
     version = "$GITDESC"
     EOF
     '';
--- a/default.nix
+++ b/default.nix
@@ -9,7 +9,25 @@ let
   excludeContent = path: typ: 
     let d = baseNameOf (toString path);
     in !(d == "guide" && typ == "directory");
-  neuronRoot = pkgs.lib.cleanSourceWith { filter = excludeContent; src = gitignoreSource ./.; };
+  projectRoot = ./.;
+  neuronSrc = pkgs.lib.cleanSourceWith { filter = excludeContent; src = gitignoreSource projectRoot; };
+  gitDescribe = pkgs.runCommand "neuron-gitDescribe" 
+    { buildInputs = [ pkgs.git ]; }
+    ''
+      mkdir $out
+      git -C ${projectRoot} describe --long --always > $out/output
+    '';
+  neuronRoot = pkgs.runCommand "neuron" { buildInputs = [ pkgs.git gitDescribe ]; }
+    ''
+    mkdir $out
+    cp -r -p ${neuronSrc}/* $out/
+    chmod -R u+w $out/
+    GITDESC=`cat ${gitDescribe}/output`
+    cat << EOF > $out/src/Neuron/Zettelkasten/Version.hs
+    module Neuron.Zettelkasten.Version where
+    version = "$GITDESC"
+    EOF
+    '';
 in {
 # Rib library source to use
   rib ? builtins.fetchTarball "https://github.com/srid/rib/archive/${ribRevision}.tar.gz"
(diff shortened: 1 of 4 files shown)


srid:
Thanks! 

This works when run via cabal:

```
$ nix-shell --run 'cabal new-run exe:neuron -- --version'
Up to date
0.1.0.0 3dcbec46b6df7d0f2d27c3847fc894a4b0984ff0
```

However the final executable, built by nix, does not show the git revision:

```
$ $(nix-build)/bin/neuron --version
0.1.0.0 UNKNOWN
```

It looks like resolving this may involve some Nix machinery in a manner similar to how [staticWhich](https://github.com/obsidiansystems/which) is used in the project.

TristanCacqueray:
It seems like `UNKNOWN` can also be the result of a missing git cli. And I'm not sure what the `gitignoreSource` is supposed to do, but it seems like it is also resulting in a 'UNKNOWN' git version string.

I'm not familiar with nix, but i was able to do fix the `nix-build` use-case with: d2dd530. Though I guess this is not correct. Is there a way to make the nix-build test faster by avoiding full rebuild?

srid:
> Though I guess this is not correct. Is there a way to make the nix-build test faster by avoiding full rebuild?

I have not researched this in full, but one idea that comes to my mind is to run `git describe` in Nix expressions, and then pass its output via an environment variable, so that Haskell sources would read them via [TH](http://hackage.haskell.org/package/th-env).

srid:
That, or simply ... a `Version.hs` module could be generated by Nix in `neuronRoot` and then `import Version (version)` in the code. If `Version.hs` doesn't change, Nix shouldn't recompile from scratch.

TristanCacqueray:
Thank you for the suggestion. Unfortunately I wasn't able to make ee3e561 works.

srid:
@TristanCacqueray Try this:

```nix
diff --git a/default.nix b/default.nix
index 34ae687..96f5084 100644
--- a/default.nix
+++ b/default.nix
@@ -29,13 +29,14 @@ let
 in import rib { 
     inherit root name additional-packages; 
     source-overrides = {
-      neuron = neuronRoot;
       # Until https://github.com/obsidiansystems/which/pull/6 is merged
       which = builtins.fetchTarball "https://github.com/srid/which/archive/5061a97.tar.gz";
       # Generate git version because gitignore prevent using gitRev
-      neuronVersion = pkgs.runCommand "version" { buildInputs = [ pkgs.git ]; }
+      neuron = pkgs.runCommand "neuron" { buildInputs = [ pkgs.git ]; }
       ''
-      cat << EOF > ${neuronRoot}/src/Neuron/Zettelkasten/Version.hs
+      mkdir $out
+      cp -r $neuronRoot/* $out/
+      cat << EOF > $out/src/Neuron/Zettelkasten/Version.hs
       module Neuron.Zettelkasten.Version where
       version = $(git describe --long --always)
       EOF
```

srid:
Oh wait, Haskell still couldn't locate Version.hs ...

srid:
@TristanCacqueray Try this: https://github.com/TristanCacqueray/neuron/pull/1

TristanCacqueray:
Alright, that worked. Though I find this a bit convoluted, it seems like the root cause is gitRev not pulling the git cli and the gitignoreSource that somehow hides the .git dir. Perhaps we could fix that and avoid using such workaround for nix-build?

In any case thank you for the guidance, this has been a good learning opportunity!

srid:
Great!

> Though I find this a bit convoluted, it seems like the root cause is gitRev not pulling the git cli and the gitignoreSource that somehow hides the .git dir.

Yes, but the ignoring of `.git` dir is for a good reason. Otherwise, `nix-build` would do a rebuild, creating a new derivation, every time anything under `.git` changes. Our approach here is actually better, and it is in line with Nix's purely function package management philosophy. We create a derivation `git-Describe` that holds the git revision (which will be rebuilt when our sources or `.git` changes), and then we pass it as a dependency to the actual neuron Haskell derivation. Now, if every time `.git` changes, only the `git-Describe` derivation will be rebuilt (and it is pretty quick; just runs `git describe`), but our Haskell derivation won't (unless the git revision changes).

PR #63, opened 2020-04-05
Add math support documentation in guide

felko:
Add documentation about mathjax support from zettel markdown.
This is needed because MMark uses unconventional notation for LaTeX input.

Some features also have unexpected bugs as `$$ ... $$` blocks fail to escape some markdown features, in particular superscript and subscript (todo: figure out if these come from mmark-ext or from mathjax directly).


srid:
Thanks! I tweaked it a bit. Deployed: https://neuron.srid.ca/2013701.html

Issue #48, opened 2020-04-02
Display tags in HTML interface


srid:
In the individual zettel pages.

PR #69, opened 2020-04-07
Display the zettel tags

felko:
Implement #48


srid:
Thanks; merged with a minor refactor.

PR #79, opened 2020-04-10
CLI revamp

### LINKED ISSUE #76: Feature Request: Add host option to `rib serve`
It would be nice to be able to access the http-server from a different computer than the one where the server was started.  

What I want to accomplish is: 

    neuron ZDIR rib serve --port 8888 --host 0.0.0.0

Thanks.

### LINKED ISSUE #77: Make rib commands fail when the zettelkasten doesn't exist
When trying to generate the site from a nonexistent zettelkasten with `neuron nonexistent-dir rib generate`, shake will create `nonexistent-dir/.shake` and two files `nonexistent-dir.output/index.html` and `nonexistent-dir.output/z-index.html` will be generated.

It would be preferable to fail in the case where the zettelkasten doesn't exist, e.g. for editor support, to notify that the command has failed instead of generating the webapp from nowhere.

I notice however that `neuron nonexistent-dir new "title"` fails with `getDirectoryContents:openDirStream: does not exist (No such file or directory)`. Maybe the existence of the zettelkasten should be checked in all cases, and provide an pretty error message instead of a Haskell error.

### LINKED ISSUE #78: Add a `.neuron` directory
Add a `.neuron` directory and 

1. move `.shake` under it, as well as  
2. move the `notes.output` directory under it.

So:

```
./notes  # zettelkasten
./notes/.neuron/shake # shake database
./notes/.neuron/output # Generated HTML files
```

Ref: https://funprog.srid.ca/neuron/selecting-the-output-html-destination.html

srid:
- [x] Advance to newer rib with revamped CLI
- [x] Resolves #78 
- [x] Resolves #76 
- [x] Resolves #77
- [x] Change CLI to be of the form: `neuron -d ~/zettelkasten <cmd>`
- [x] Handle rib's cli args
  - [x] `neuron -d ~/zettelkasten rib -ws :8080` for serving (discard the `s :8080` to watch and generate only).
  - [x] `neuron -d ... rib --output=/tmp/output` to override output directory
- [x] Refactor CLI code
- [x] Add `neuron open` to open the locally generated output, using `xdg-open` on Linux and `open` on mac.
- [x] Update docs & test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ dist-*
 result
 .shake
 
-guide.output
+guide/.neuron
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Styling changes:
   - Tags are restyled and positioned below
   - Produce compact CSS in HTML head
+- CLI revamp
+  - Zettelkasten directory is now provided via the `-d` argument.
+    - Its default, `~/zettelkasten`, is used when not specified.
+    - This directory must exist, otherwise neuron will error out.
 
 ## 0.2.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
   - Zettelkasten directory is now provided via the `-d` argument.
     - Its default, `~/zettelkasten`, is used when not specified.
     - This directory must exist, otherwise neuron will error out.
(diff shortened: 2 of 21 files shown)


srid:
Documentation changes: https://github.com/srid/neuron/commit/9f40e33cb16994817620c99d80173328d8666170

Issue #59, opened 2020-04-04
More convenient links


Nadrieril:
Linking between zettels is arguably one of the main features of the method, yet the syntax is quite clunky: `[20200101](z://)` (especially when typing on a phone). Would it be possible to make this more convenient ? For example, having just `[20200101]` seems quite markdown-y, since markdown supports [link reference definitions](https://spec.commonmark.org/0.12/#link-reference-definition).

Nadrieril:
I see this is a limitation of mmark :/ I would support switching to pandoc, because it would allow so much more stuff like syntax highlighting, math or latex

srid:
Having just `[20200101]` was what I was originally trying to do. Yea, it is a limitation of mmark, but maybe we can extend mmark or mmark-ext to support it.

Syntax highlighting is already supported by mmark/neuron. See [example](https://haskell.srid.ca/2013603.html). You need to include the language name as the first in the code fence (see [source](https://raw.githubusercontent.com/srid/haskell-zettelkasten/89235bcb8e0a94c9e07993ae5434976853689023/notes/2013603.md) for that example).

Math supported was added in [this PR](https://github.com/srid/neuron/pull/16). And I'm not sure how you plan to use latex in a Zettelkasten.

I feel ambivalent about switching to Pandoc; it will require some thought. MMark is just simpler to deal with. But if it proves to be too restrictive, and Pandoc is not too complex to use, then it may be worth considering.

Nadrieril:
Oh, I had no idea neuron supported highlighting and maths, that's quite cool ! FWIW I've used Pandoc to parse markdown before and found it relatively straightforward.
I just discovered the `<https://example.com>` markdown syntax, which seems to be equivalent to `[https://example.com](https://example.com)`. mmark supports it. Maybe we could convert `<20200101>` and `<z://20200101>` to zettel links ?

srid:
Interesting, I didn't know about the angle brackets. 

> Maybe we could convert `<20200101>` and `<z://20200101>` to zettel links ?

Yea, I'd be open to that. The exact format will have to be decided first. The ID in "z://20200101" is currently ignored by Neuron, which requires the ID to be in link text. So it is more straightforward to support `<20200101>`.

But in general I wonder if neuron should support `z://<id>` style links as well.

srid:
One idea is to use `z:/` to refer to _all_ link types (deprecating zcf, zquery, zquerycf):

```
z://link/2012342 # Equivalent to [2012342](z://...)
[...]

Issue #70, opened 2020-04-07
Standardize on zettel ID scheme


srid:
The general expectation so far is that the IDs are of [a specific format](https://neuron.srid.ca/2011403.html), although the app is pretty lenient ("index.md" for example works).

Some users, like @Nadrieril from #60, use a different ID scheme. 

One of the ideas in #59 is to support short and simple links like `<2011502>` (instead of the current `[2011502](z:/)`). Knowing what to expect ahead is necessary to properly parse links.

Should the next version of neuron:

- allow only numeric IDs like `2011502`?
- allow custom alphanumeric IDs like `20dtbvs` ([cf](https://old.reddit.com/r/Zettelkasten/comments/fa09zw/shorter_zettel_ids/fmhi8ky/))

I suppose a third option is to strongly recommend numeric IDs - but fail gracefully where users use alphanumeric IDs. For example, zquery's `theme=withDate` argument should render using the default theme, and render a warning in the HTML when the IDs are not date encoded numeric IDs. This inequality would have to be represented at the type level:

```haskell
data ZettelID 
  = DateID Day -- Numeric
  | CustomID Text -- Alpha numeric
```

Nadrieril:
As mentioned, I'm in favor of supporting custom ids because that's what I use. The third option is good enough for me.
A different argument for arbitrary ids would be to be compatible with imports from other sources. I'm thinking specifically Zotero or similar software. The idea would be that I import each reference as a zettel; that means I can use the same id in my zettels and my LaTeX documents, and I get backlinks for each reference too.

srid:
Hmm, I'm not keen on supporting importing from other software. But just out of curiosity, what is their ID formats like (if not of alphanumeric letters)?

Nadrieril:
I didn't mean include such support in neuron, but just leave it possible if I want to write a script that does it.
Zotero (and bibtex) support arbitrary ids, but by default they're alphanumeric and look something like authorTitleOfPaper2020

srid:
https://neuron.srid.ca/2011403.html

PR #80, opened 2020-04-10
Support short links

Nadrieril:
This adds support for short `<1234567>` links, as discussed in #59. In fact what this does is render any `[foo](1234567)` link as a zettel link. That can easily be changed if needed.

Some questions: where should I document this ? Do you use a Haskell formatter and if yes which one ?

### DIFF
--- a/guide/2011504.md
+++ b/guide/2011504.md
@@ -5,19 +5,20 @@ title: Linking
 To link to another zettel, use the following syntax:
 
 ```markdown
-This is a zettel file, which links to another zettel: 
+This is a zettel file, which links to another zettel:
 
-- [2008403](z:). 
+- [2008403](z:).
 
-You may also annotate the link (ignored by neuron): 
+You may also annotate the link (ignored by neuron):
 
-- [2008403](z://foo-bar). 
+- [2008403](z://foo-bar).
 ```
 
 The `z:` protocol instructs neuron to automatically create a link between the associated zettels, which ultimately affects your zettel graph. Neuron also renders the link along with the ID and title of the linked zettel.
 
-If your link is merely a reference to another zettel, and you do not wish it to be also a graph connection, use `zcf:` instead. Neuron will link the zettels, but the link would be ignored when building the [2011503](zcf://graph-view). 
+If your link is merely a reference to another zettel, and you do not wish it to be also a graph connection, use `zcf:` instead. Neuron will link the zettels, but the link would be ignored when building the [2011503](zcf://graph-view).
 
 ## Other link types
 
 * [2011506](z://zquery)
+* [2014501](z://short-links)
--- a/guide/2011504.md
+++ b/guide/2011504.md
@@ -5,18 +5,18 @@ title: Linking
 To link to another zettel, use the following syntax:
 
 ```markdown
-This is a zettel file, which links to another zettel: 
+This is a zettel file, which links to another zettel:
 
-- [2008403](z:). 
+- <2008403>.
 
-You may also annotate the link (ignored by neuron): 
+You may also annotate the link (ignored by neuron):
 
-- [2008403](z://foo-bar). 
+- [foo-bar](2008403).
 ```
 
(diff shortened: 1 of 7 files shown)


srid:
> render any [foo](1234567) link as a zettel link

Can you post a screenshot? What happens to the "foo" in HTML view?

srid:
> Do you use a Haskell formatter and if yes which one ?

ormolu. See https://github.com/srid/neuron/blob/master/CONTRIBUTING.md

Nadrieril:
> Can you post a screenshot? What happens to the "foo" in HTML view?

`[foo](1234567)`, `<1234567>` (which is the same as `[1234567](1234567)`) and `[1234567](z://)` are all rendered exactly the same. The foo is dropped entirely. I imagined this would only be temporary: I would expect `[foo](1234567)` to be a link with text `"foo"`.

Nadrieril:
How can I make tests for this new feature ? I would imagine tests that test both the html output of a file and the graph output of a bunch of files, but I don't know how to write that myself.

srid:
I think you should ignore the `[foo](1234567)` case. That can be handled in a different PR with a concrete spec.

This PR should handle the single case of `<1234567>` (same as [1234567](1234567)), making it behave the same as `[1234567](z:)`.

> How can I make tests for this new feature ? 

I've been meaning to begin writing tests for link actions, but if you want to take stab, go ahead! Tests use [hspec](https://hspec.github.io/). Maybe add a `test/Neuron/Zettelkasten/Link/ActionSpec.hs` that tests `linkActionFromLink` by running it over each individual cases. See `IDSpec.hs` for reference implementation.

srid:
> I think you should ignore the `[foo](1234567)` case. 

From the docs in the diff,
```
You may also annotate the link (ignored by neuron):

- [2008403](z://foo-bar). 	- [foo-bar](2008403).
```

I like the annotation feature in general. But if we finalize the spec, then we can't expect it to be a link with text "foo-bar" (that would not be backwards compatible). 

I suppose the user could specify the the link should be rendered normally using a query like this:

```
[foo-bar](2008403?linkTheme=normal)
```

guide/2011504.md
```diff
@@ -5,18 +5,18 @@ title: Linking
 To link to another zettel, use the following syntax:
 
 ```markdown
-This is a zettel file, which links to another zettel: 
+This is a zettel file, which links to another zettel:
 
-- [2008403](z:).
```
srid, line 10: Until the _full_ range of links (z:,  zcf:, zquery:, zcfquery:) get their alias syntax in neuron, docs should continue to recommend the old style. You can mention as an aside that "<1234567>" is an alternative syntax (with a note that it would be expanded as discussed [here](https://github.com/srid/neuron/issues/59)) for z:.

Nadrieril:
> I like the annotation feature in general.

What's the use for annotations ? Is it to make it easier to know which zettel it corresponds to ? Then I'd say `<1234567?comment=foo-bar>`, makes a lot more sense semantically than ignoring the link text.
But more generally, for a thing that kinda functions as a wiki, it probably makes sense to sometimes use human-readable ids.

> if you want to take stab, go ahead!

I'll let you do it, you know the codebase. I'll happily add tests to it once it's started though.

srid:
Yes, annotations are mainly to act as an immediate visual hint when editing a zettel to know what a particular zettel ID refers to at first glance without having to open it.

I prefer something like `<1234567#foo-bar>` (i.e., the "anchor" is ignored) for annotations. Although anchors could potentially be used to link to a particular section of the zettel. But this can happen in different PR I guess. 

> But more generally, for a thing that kinda functions as a wiki, it probably makes sense to sometimes use human-readable ids.

The proposal at #70 currently stands at supporting alphanumeric IDs. Perhaps we could add extra characters, like `-`, to it - so that human-readable ids like "foo-bar.md" is supported?

> I'll let you do it, you know the codebase. I'll happily add tests to it once it's started though.

Alright.

srid:
Added tests and merged in https://github.com/srid/neuron/commit/0aac0fabeb754364cde18762537cc9bcd6fb689b

Issue #88, opened 2020-04-11
Link tags to a zettel that queries it


srid:
Tags are currently shown as a ribbon at the bottom right of zettels.
![image](https://user-images.githubusercontent.com/3998/79053256-fb3f4900-7c09-11ea-9511-8c3bcdae326b.png)

These tags should be made clickable. Clicking them should go to the zettel that uses `zquery` to query that particular tag. If there are more than one such zettel, then pick a suitable one (one that exclusively uses this tag, and probably created the earliest in time).

srid:
Currently this goes to search.html, which is indeed an improvement. I'd also like somewhat to access zettels that query the tag, but that's a different feature.

PR #90, opened 2020-04-11
Web search

felko:
This implements a search page where users can query the zettelkasten by title and by tags.

- [x] Search zettels by title
- [x] Zettel tags redirect to the search page where one can see all zettels with that tag (alternative to #88 )
- [x] Advanced search
  - [x] Make a separate field for selecting tags explicitly
  - [ ] ~~Filter by date~~ Will integrate dates into the index later
  - [ ] ~~Sorting (alphabetically, chronologically, ...)~~ Useless without the date and more fields to sort by

I was planning on using Semantic UI modules for the interface of advanced search, which requires JavaScript content. We therefore must think about whether or not the search page should be accessible locally (or even offline) which would mean adding an npm dependency.


srid:
Just took a casual look. This is great! 

> I was planning on using Semantic UI modules for the interface of advanced search, which requires JavaScript content. We therefore must think about whether or not the search page should be accessible locally (or even offline) which would mean adding an npm dependency.

That should be fine, as long as the new dependencies are managed by Nix. 

As for offline, that's not even supported right now (see `src-bin/Main.hs` which is pulling semantic UI css from CDN) - but we probably should (i.e., copy over semantic ui files using `Rib.buildStaticFiles`) .

src/Neuron/Zettelkasten/Route.hs
```diff
@@ -27,14 +28,27 @@ import qualified Text.URI as URI
 data Route store graph a where
   Route_Redirect :: ZettelID -> Route ZettelStore ZettelGraph ZettelID
   Route_ZIndex :: Route ZettelStore ZettelGraph ()
+  Route_Search :: {searchTerms :: Maybe String, searchTags :: [Text]} -> Route ZettelStore ZettelGraph ()
```
srid, line 31: This looks semantically off. The constructor arguments should contain only the information needed to determine the route .html filename. As this is `search.html` in this case, there should be no constructor arguments at all. Now, the "data" required to *render* the route is to be specified in the GADT type parameters (`store`, `graph` and `a`). Looking at `renderSearch`, it doesn't look like you need anything extra to render the search page. So this should really be:

```haskell
Route_Search :: Route ZettelGraph ZettelGraph ()
```

src/Neuron/Zettelkasten/Route.hs
```diff
@@ -27,14 +28,27 @@ import qualified Text.URI as URI
...
+
 instance IsRoute (Route store graph) where
   routeFile = \case
     Route_Redirect zid ->
       routeFile $ Route_Zettel zid
     Route_ZIndex ->
       pure "z-index.html"
+    Route_Search {..} -> do
+      query <- renderSearchQuery searchTerms searchTags
+      pure $ "search.html" ++ query
```
srid, line 51: See above; this should just be `pure "search.html"`

src/Neuron/Zettelkasten/View.hs
```diff
@@ -39,6 +41,9 @@ import qualified Rib.Parser.MMark as MMark
 import Text.MMark (useExtensions)
 import Text.Pandoc.Highlighting (styleToCss, tango)
 
+searchScript :: Text
+searchScript = $(embedStringFile "./src-script/neuron-web-search/search.js")
```
srid, line 45: This file should be placed in `./src-js/search.js`. 

`src-script` is for shell scripts.

src/Neuron/Zettelkasten.hs
```diff
@@ -106,12 +107,20 @@ generateSite config writeHtmlRoute' zettelsPat = do
...
   -- Write alias redirects, unless a zettel with that name exists.
   aliases <- Z.getAliases config zettelStore
   forM_ aliases $ \Z.Alias {..} ->
     writeHtmlRoute targetZettel (Z.Route_Redirect aliasZettel)
   pure (zettelStore, zettelGraph)
 
+writeIndex :: Z.ZettelStore -> Action ()
+writeIndex store = do
+  let results = Z.runQuery store []
+  Rib.writeFileCached "index.json" $ decodeUtf8 $ Aeson.encode (Aeson.toJSON results)
```
srid, line 122: `toString $ Data.Aeson.Text.encodeToLazyText results`?

srid:
A couple of comments regarding the footer:

- The search icon should be placed before the z-index tree icon (which remains at the end)
- The icons are not aligned with equal gaps (unless editUrl is enabled). See screenshot

image

Here, the search icon should be at the center, with the z-index icon placed in the far right. If `editUrl` is enabled, all the four icons should be spaced equidistantly.

src/Neuron/Zettelkasten/View.hs
```diff
@@ -144,7 +173,11 @@ renderTags tags = do
     -- the top pushes the zettel content down, introducing unnecessary white
     -- space below the title.
     span_ [class_ "ui black right ribbon label", title_ "Tag"] $ do
-      toHtml @Text tag
+      a_
+        [ href_ (Rib.routeUrlRel $ Route_Search Nothing [tag]),
```
srid, line 177: Oh, and here it would look like:
```
Rib.routeUrlRel Route_Search <> "?tag=" <> tag ...
```
(Would be nice to add a function to Rib.Route that returns something like `Text.URI` instead).

srid:
Also, search doesn't work in the local version (`neuron open`). Maybe embedding the JSON in search.html is a solution ...

PR #93, opened 2020-04-12
Nix derivation: accept pkgs as an arg

ceedubs:
This allows declarative installations to pin the nixpkgs version to
enable deterministic builds and overrides.


srid:
Using the same `pkgs` as that of [rib](https://github.com/srid/rib)'s default.nix would be ideal, but this works for now. :+1:

PR #100, opened 2020-04-13
feat(website): Add MMark extension to render table with styles

### LINKED ISSUE #91: Styles questions/suggestions
Are you open to change some of the styles for zettles?

If so, (IMHO) there are some styles that could be improved.

**Use [Semantic UI classes for tables](https://semantic-ui.com/collections/table.html)** (`ui celled table`)

Without the classes:
Screenshot 2020-04-12 at 11 43 51

With the classes:
Screenshot 2020-04-12 at 11 43 34

**Full width code blocks**

Code blocks with not language take the full with, but the ones with code highlight take 80% ([I know this is intentional](https://github.com/srid/neuron/blob/4f42faf05bf1213403152cd1db9d6abd96aa3c35/src/Neuron/Zettelkasten/View.hs#L253-L258)).

This would be aligned with GitHub styles for code blocks. To me it feels more natural for code blocks to take the full width.

Current:

Screenshot 2020-04-12 at 11 47 25

Proposed:

Screenshot 2020-04-12 at 11 47 45

gillchristian:
Fixes #91 

This PR adds a MMark extension to render tables with classes.

### TODO

- [ ] Use Semantic UI classes for cell alignment?
- [ ] ...

#### Before

Screenshot 2020-04-13 at 19 53 57

#### After

Screenshot 2020-04-13 at 19 53 42

### DIFF
--- a/src/Neuron/Zettelkasten/Markdown/Extension.hs
+++ b/src/Neuron/Zettelkasten/Markdown/Extension.hs
@@ -1,0 +1,48 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Neuron.Zettelkasten.Markdown.Extension
+  ( customRender,
+  )
+where
+
+import Control.Monad (forM_)
+import Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NE
+import Lucid
+import Text.MMark.Extension (Block (Table), Extension)
+import qualified Text.MMark.Extension as Ext
+
+newline :: Html ()
+newline = "\n"
+
+-- https://github.com/mmark-md/mmark/blob/8f5534d8068c2b7a139b893639ee5920bcaedd84/Text/MMark/Render.hs#L111-L136
+customRender :: Extension
+customRender = Ext.blockRender $ \old block ->
+  case block of
+    (Table calign (hs :| rows)) -> do
+      table_ [class_ "ui celled table"] $ do
+        newline
+        thead_ $ do
+          newline
+          tr_
+            $ forM_ (NE.zip calign hs)
+            $ \(a, h) ->
+              th_ (alignStyle a) (snd h)
+          newline
+        newline
+        tbody_ $ do
+          newline
+          forM_ rows $ \row -> do
+            tr_
+              $ forM_ (NE.zip calign row)
+              $ \(a, h) ->
+                td_ (alignStyle a) (snd h)
+            newline
+        newline
+      newline
+    other -> old other
+  where
+    alignStyle Ext.CellAlignDefault = []
+    alignStyle Ext.CellAlignLeft = [style_ "text-align:left"]
+    alignStyle Ext.CellAlignRight = [style_ "text-align:right"]
+    alignStyle Ext.CellAlignCenter = [style_ "text-align:center"]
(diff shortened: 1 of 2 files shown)


src/Neuron/Zettelkasten/Markdown/Extension.hs
```diff
@@ -0,0 +1,48 @@
...
+                td_ (alignStyle a) (snd h)
+            newline
+        newline
+      newline
+    other -> old other
+  where
+    alignStyle Ext.CellAlignDefault = []
+    alignStyle Ext.CellAlignLeft = [style_ "text-align:left"]
+    alignStyle Ext.CellAlignRight = [style_ "text-align:right"]
+    alignStyle Ext.CellAlignCenter = [style_ "text-align:center"]
```
srid, line 48: Is our only choice here to fork [upstream code](https://github.com/mmark-md/mmark/blob/8f5534d8068c2b7a139b893639ee5920bcaedd84/Text/MMark/Render.hs#L111-L136) just to modify the class name? Is there a way to re-use upstream code without having to fork it?

src/Neuron/Zettelkasten/Markdown/Extension.hs
```diff
@@ -0,0 +1,48 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Neuron.Zettelkasten.Markdown.Extension
+  ( setTableClass,
+  )
+where
+
+import qualified Data.List.NonEmpty as NE
```
gillchristian, line 9: [Q] I did not find a Relude equivalent to `NE.zip` so I kept this one

src/Neuron/Zettelkasten/Markdown/Extension.hs
gillchristian: This is a good point. I opened the PR in this state to get the conversation started but I'm not happy to fork the code either.

I took a (quick) look at mmark and didn't find a way to do it. But extending the way certain blocks are rendered seems common enough and could be covered by mmark I guess.

gillchristian: > extending the way certain blocks are rendered seems common enough and could be covered by mmark I guess

To be precise, it is supported by `Ext.blockRender`. But it would be great to not have to copy the whole rendering of the table just to add a class. I guess that's what you also meant.

src/Neuron/Zettelkasten/Markdown/Extension.hs
srid: That's fine, yea.

src/Neuron/Zettelkasten/Markdown/Extension.hs
srid: Mark's [suggestion](https://github.com/mmark-md/mmark/issues/70#issuecomment-613639892) to use [`with`](https://hackage.haskell.org/package/lucid-2.9.12/docs/Lucid.html#v:with) seems like it would obviate the need to fork the renderer.

gillchristian: I'll try that. 👌

gillchristian: Much better now!

PR #102, opened 2020-04-15
ZettelStore: more helpful lookup failure message

ceedubs:
I made a typo and transposed a couple of numbers in a zettel link. I was
getting an error message like `No such zettel: ZettelDateID 2020-03-19`,
but since there were a bunch of links in the doc, it took me a while to
figure out that this corresponded to the `[2011401]` link in my doc.
I've changed the error message for date IDs to include the text as it
would actually appear in the link in addition to the parsed day/index.

It would probably also be nice to get source line numbers on error
messages, but that sounds like a more ambitious effort.

I've never really written Haskell, so if I'm doing anything weird, it's
for no good reason, and please let me know.

### DIFF
--- a/test/Neuron/Zettelkasten/StoreSpec.hs
+++ b/test/Neuron/Zettelkasten/StoreSpec.hs
@@ -1,0 +1,29 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Neuron.Zettelkasten.StoreSpec
+  ( spec,
+  )
+where
+
+import Control.Exception (evaluate)
+import qualified Data.Map.Strict as Map
+import Data.Time.Calendar
+import qualified Neuron.Zettelkasten.ID as Z
+import qualified Neuron.Zettelkasten.Store as Z
+import Relude
+import Test.Hspec
+
+spec :: Spec
+spec = do
+  describe "Zettel Store" $ do
+    let day = fromGregorian 2020 3 19
+        emptyStore = Map.empty
+        dateZid = Z.ZettelDateID day 1
+        customZid = Z.ZettelCustomID "custom-id"
+    context "nonexistent zettel ID lookup" $ do
+      it "errors with text representation of date zettel ID" $ do
+        evaluate (Z.lookupStore dateZid emptyStore) `shouldThrow` errorCall "No such zettel: 2011401"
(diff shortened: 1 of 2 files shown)


srid:
Ah, nice. You should change the error message to just be like:

```
No such zettel: 2011401
``` 

(Showing the date of the zettel is not important). 

> It would probably also be nice to get source line numbers on error messages, but that sounds like a more ambitious effort.

Agreed. Could you open an issue for it?

test/Neuron/Zettelkasten/StoreSpec.hs
```diff
@@ -0,0 +1,29 @@
...
+  describe "Zettel Store" $ do
+    let day = fromGregorian 2020 3 19
+        emptyStore = Map.empty
+        dateZid = Z.ZettelDateID day 1
+        customZid = Z.ZettelCustomID "custom-id"
+    context "nonexistent zettel ID lookup" $ do
+      it "errors with date zettel ID date and text" $ do
+        evaluate (Z.lookupStore dateZid emptyStore) `shouldThrow` errorCall "No such zettel: ZettelDateID 2020-03-19 1 (2011401)"
+      it "errors with custom zettel ID text" $ do
+        evaluate (Z.lookupStore customZid emptyStore) `shouldThrow` errorCall "No such zettel: ZettelCustomID \"custom-id\""
```
srid, line 29: Custom zids should show an error like:

```
No such zettel: custom-id
```

(To be consistent with date zids)

ceedubs:
Sounds good. I can probably get to this tomorrow.

ceedubs:
Feedback addressed and issue #104 created.

srid:
Thanks, and welcome to Haskell development! :smiley:

ceedubs:
> welcome to Haskell development! 😃

:) thank you. I've written a lot of functional Scala and a bit of OCaml so it's not as strange to me as it would be to many. For quite a while it's been the language that I'd like to use if more companies were hiring for it 😅

I noticed #103, and I do think that Nix has a very high barrier and will deter many potential users (though homebrew might be more of a solution to that than Stack). However, I wanted to let you know that as someone who _does_ use Nix and doesn't use Haskell, contributing to this project was delightfully smooth. I was able to follow the instructions in the README to drop into a `nix-shell` (directly or via one of the `bin/` scripts) with the necessary environment, and compiling and running tests automatically watched for source changes. If I start doing more Haskell development I'll want to figure out how to set up a proper language server, but Vim on top of the great Nix setup that you have here worked surprisingly well!

Issue #113, opened 2020-04-17
Feature Request: Add builtin optional metadata: author and date


michelk:
I would be nice if the author and the date of last change would be rendered small at the bottom of the page.

srid:
Note that the date of last change is different from the zettel date (which is date of its creation). It is indeed a good feature to include the last modified date in the footer.

I'm curious about author, though. Do you have you a zettelkasten with multiple authors? There is already an author field in configuration (neuron.dhall); it is used only in OpenGraph meta tags.

michelk:
For community zettelkästen or enterprise knowledge bases this would be
a helpful feature.

michelk:
What do you think about rendering all (custom) meta-data entries at the footer of the page?

srid:
> What do you think about rendering all (custom) meta-data entries at the footer of the page?

That may not be a bad idea. But I'd be wary of rendering *all* custom meta-data by default. Not sure if that's desirable.

If author is the only thing you'd like to be rendered, why not make it a tag like:

```
title: My zettel
tags:
  - author/michelk
```

?

With the upcoming #121 you could do nifty things like show a list of all authors, with a link to a page that shows all zettels by them.

michelk:
Sounds like a good idea.

srid:
Date of zettel creation is now displayed at the bottom, in 8acef9594a4edc02b4a828df5d54bf587f2d0395

PR #114, opened 2020-04-17
optional meta-data fields 'author' and 'date'

michelk:
Addresses #113


michelk:
I agree, the date field is not well defined.
If we take the file status of last modification this could get corrupted by git checkouts; maybe it is better to leave it out. 

 Currently, I don't know what to do in `routeOpenGraph`.

Issue #49, opened 2020-04-02
Hierarchical tags support


srid:
If a zettel has the following metadata
```markdown
---
title: Some zettel
tags:
  - journal/work
---
```

Should querying for `tag=journal` include `journal/work` zettels as well? Or would this behaviour be specified explicitly by the user somehow?

srid:
Some discussion about this [at Zulip](https://funprog.srid.ca/neuron/hierarchical-tags.html).

srid:
Here's a proposal for how this would look: https://funprog.srid.ca/neuron/hierarchical-tags.html#194321595

srid:
Need to document, once the spec/API/etc. are finalized to be our satisfaction.

srid:
Documented: https://neuron.zettel.page/535407ad.html

PR #115, opened 2020-04-17
Hierarchical tags

felko:
This PR implements hierarchical tags as described in #49.

Tags can be nested using the same syntax as POSIX file paths, and queries support tag globbing as well (including recursive globs).


src/Neuron/Web/View.hs
```diff
@@ -55,11 +56,11 @@ searchScript = $(embedStringFile "./src-js/search.js")
 helloScript :: Text
 helloScript = $(embedStringFile "./src-purescript/hello/index.js")
 
-mkSearchURI :: MonadThrow m => Maybe Text -> [Text] -> m URI
+mkSearchURI :: MonadThrow m => Maybe Text -> [Tag] -> m URI
```
srid, line 59: What's this `Maybe Text`? If it is not used, it should be removed entirely.

src/Neuron/Zettelkasten/Tag.hs
```diff
@@ -0,0 +1,54 @@
...
+  )
+where
+
+import Data.Aeson
+import Relude
+import Relude.Extra.Foldable1
+import System.FilePath
+import System.FilePattern as FilePattern
+
+newtype Tag = Tag {tagToText :: Text}
```
srid, line 28: `tagToText` should be renamed to `unTag`.

src/Neuron/Zettelkasten/Tag.hs
```diff
@@ -0,0 +1,54 @@
...
+  deriving (Eq, Ord, Show, ToJSON, FromJSON)
+
+instance Semigroup Tag where
+  Tag t <> Tag t' = Tag (toText $ toString t </> toString t')
+  sconcat tags = Tag (toText $ foldl1' (</>) $ fmap (toString . tagToText) tags)
+
+newtype TagPattern = TagPattern {toFilePattern :: FilePattern}
```
srid, line 35: `unTagPattern`

src/Neuron/Zettelkasten/Tag.hs
```diff
@@ -0,0 +1,54 @@
...
+  deriving (Eq, Show)
+
+instance Semigroup TagPattern where
+  TagPattern p <> TagPattern p' = TagPattern (p </> p')
```
srid, line 39: Use `GeneralizedNewtypeDeriving`?

src/Neuron/Zettelkasten/Tag.hs
```diff
@@ -0,0 +1,54 @@
...
+import Relude.Extra.Foldable1
+import System.FilePath
+import System.FilePattern as FilePattern
+
+newtype Tag = Tag {tagToText :: Text}
+  deriving (Eq, Ord, Show, ToJSON, FromJSON)
+
+instance Semigroup Tag where
+  Tag t <> Tag t' = Tag (toText $ toString t </> toString t')
+  sconcat tags = Tag (toText $ foldl1' (</>) $ fmap (toString . tagToText) tags)
```
srid, line 33: Let's use `</>` directly. No need for this instance.

src/Neuron/Zettelkasten/Tag.hs
```diff
@@ -0,0 +1,54 @@
...
+
+newtype TagPattern = TagPattern {toFilePattern :: FilePattern}
+  deriving (Eq, Show)
+
+instance Semigroup TagPattern where
+  TagPattern p <> TagPattern p' = TagPattern (p </> p')
+
+tagPatternToText :: TagPattern -> Text
+tagPatternToText = toText . toFilePattern
```
srid, line 42: This should be removed (in one place you are converting back to string anyway).

src/Neuron/Zettelkasten/Tag.hs
```diff
@@ -0,0 +1,54 @@
...
+literalPattern = TagPattern . toString . tagToText
+
+tagMatch :: TagPattern -> Tag -> Bool
+tagMatch (TagPattern pat) (Tag tag) = pat ?== toString tag
+
+isSubTag :: Tag -> Tag -> Bool
+isSubTag tag tag' = tagMatch (literalPattern tag <> TagPattern "**") tag'
+
+isStrictSubTag :: Tag -> Tag -> Bool
+isStrictSubTag tag tag' = tagMatch (literalPattern tag <> TagPattern "*" <> TagPattern "**") tag'
```
srid, line 54: These 3 functions (not `tagMatch`) aren't used anywhere; and should be removed.

test/Neuron/Zettelkasten/TagSpec.hs
```diff
@@ -0,0 +1,53 @@
...
+      "project/*/task",
+      "project/neuron/hierarchical-tags/task",
+      False
+    ),
+    ( "can match empty with globstar",
+      "math/**/note",
+      "math/note",
+      True
+    )
+  ]
```
srid, line 53: Could you add the following as well?

* `journal`
* `journal/*/**`

src/Neuron/Zettelkasten/Query.hs
```diff
@@ -44,13 +47,13 @@ parseQuery uri =
   flip mapMaybe (URI.uriQuery uri) $ \case
     URI.QueryParam (URI.unRText -> key) (URI.unRText -> val) ->
       case key of
-        "tag" -> Just $ ByTag val
+        "tag" -> pure $ ByTag (TagPattern $ toString val)
```
srid, line 50: nitpick: Change back to `Just` please. :-)

test/Neuron/Zettelkasten/TagSpec.hs
```diff
@@ -0,0 +1,53 @@
...
+
+shouldMatch :: Z.TagPattern -> Z.Tag -> Expectation
+shouldMatch pat tag
+  | Z.tagMatch pat tag = pure ()
+  | otherwise = expectationFailure $ toString $ Z.tagPatternToText pat <> " was expected to match " <> Z.tagToText tag <> " but didn't"
+
+shouldNotMatch :: Z.TagPattern -> Z.Tag -> Expectation
+shouldNotMatch pat tag
+  | Z.tagMatch pat tag = expectationFailure $ toString $ Z.tagPatternToText pat <> " wasn't expected to match tag " <> Z.tagToText tag <> " but it did"
+  | otherwise = pure ()
```
srid, line 23: These functions should be moved to the end of the file (`spec` should appear at the top).

test/Neuron/Zettelkasten/TagSpec.hs
srid: I feel like we should test for both shoud and shouldNot match for each pattern.

Viz.:

```haskell
cases :: [(String, String, [Text], [Text])]
cases =
  [ ( "matches a single component",
      "project/*/task",
      ["project/foo/task", "project/bar-baz/task"],
      ["project", "project/foo", "project/foo/bar/baz"]
      ),
    ...
  ]
```

src/Neuron/Zettelkasten/Tag.hs
srid: Oh wait, let's just use `</>` directly instead.

src/Neuron/Web/View.hs
felko: That was in the web search PR, but it avoids URLs like `search.html?q=&tag=science` which look ugly. Instead the `q` query parameter is not included at all in the generated URL.

src/Neuron/Zettelkasten/Tag.hs
felko: Do you mean defining `type Tag = FilePath` directly?

src/Neuron/Web/View.hs
srid: There is only one place where it is called, and with a value of `Nothing` (never a `Just` value). What would happen if you just remove this argument? (And I never saw any "q" parameter in neuron; where is it used actually?).

srid: Neuron compiles fine with this change: 

```diff
diff --git a/src/Neuron/Web/View.hs b/src/Neuron/Web/View.hs
index f602dfd..bd7cea3 100644
--- a/src/Neuron/Web/View.hs
+++ b/src/Neuron/Web/View.hs
@@ -55,13 +55,12 @@ searchScript = $(embedStringFile "./src-js/search.js")
 helloScript :: Text
 helloScript = $(embedStringFile "./src-purescript/hello/index.js")
 
-mkSearchURI :: MonadThrow m => Maybe Text -> [Text] -> m URI
-mkSearchURI terms tags = do
+mkSearchURI :: MonadThrow m => [Text] -> m URI
+mkSearchURI tags = do
   let mkParam k v = URI.QueryParam <$> URI.mkQueryKey k <*> URI.mkQueryValue v
-      qParams = maybeToList (fmap (mkParam "q") terms)
       tagParams = fmap (mkParam "tag") tags
   route <- URI.mkPathPiece (Rib.routeUrlRel Route_Search)
-  params <- sequenceA (qParams ++ tagParams)
+  params <- sequenceA tagParams
   pure
     emptyURI
       { uriPath = Just (False, [route]),
@@ -69,14 +68,14 @@ mkSearchURI terms tags = do
       }
 
 -- TODO: render error message when the query is invalid
-mkSearchQuery :: Maybe Text -> [Text] -> Text
-mkSearchQuery terms tags =
+mkSearchQuery :: [Text] -> Text
+mkSearchQuery tags =
   fromMaybe
     (Rib.routeUrlRel Route_Search)
-    (URI.render <$> mkSearchURI terms tags)
+    (URI.render <$> mkSearchURI tags)
 
 mkSingleTagQuery :: Text -> Text
-mkSingleTagQuery tag = mkSearchQuery Nothing [tag]
+mkSingleTagQuery tag = mkSearchQuery [tag]
 
 renderRouteHead :: Monad m => Config -> Route store graph a -> store -> HtmlT m ()
 renderRouteHead config r val = do
```

srid: I'll do this change later on `master`. Nevermind. Let's keep this PR focused.

felko: Alright, not really related to the PR but I'll do it 👍

src/Neuron/Zettelkasten/Tag.hs
srid: Let's take a step back and ask: where is it that you need to concatenate two tag patterns? If there is no need for it, let's eliminate this instance entirely.

felko: Right, this isn't needed anymore if I remove `isSubTag` and `isStrictSubTag` 👍

src/Neuron/Zettelkasten/Query.hs
```diff
@@ -1,8 +1,10 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedLists #-}
```
srid, line 3: Is this used?

test/Neuron/Zettelkasten/QuerySpec.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE OverloadedLists #-}
```
srid, line 1: Here as well.

test/Neuron/Zettelkasten/Link/ActionSpec.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE OverloadedLists #-}
```
srid, line 1: Here.

felko:
While writing more tests I noticed that trailing `/` might cause problems For example, `journal` did not match `journal/` and `journal/*/**` matches `journal/`. Should tags an

test/Neuron/Zettelkasten/QuerySpec.hs
srid: In fact it is used in a number of places.

felko:
#

test/Neuron/Zettelkasten/ZettelSpec.hs
```diff
@@ -21,12 +23,12 @@ spec = do
     let day = fromGregorian 2020 3 19
         zid = ZettelDateID day 1
         dummyContent = either error id $ parsePure "<spec>" "Dummy"
-        zettel = Zettel zid "Some title" ["science"] dummyContent
+        zettel = Zettel zid "Some title" [Tag "science"] dummyContent
```
srid, line 26: Could you add an another tag specific to this PR here? 

```
[Tag "science", Tag "science/biology"]
```

srid:
"journal/" is not a sensible tag (why would anyone add a slash at the end?). We don't have to support it.

Issue #110, opened 2020-04-16
Add `neuron query --list-tags`


srid:
A way to retrieve the list of all tags used in a Zettelkasten.

CLI spec: `(--list-tags | --tag TAG | --uri)`

srid:
As well as perhaps `--uri zquery://tags` (so that we can have a zettel show a list of links to tags as well).

felko:
Since #115, a `--tag-tree` command might be helpful to allow editors plugins to implement some kind of "tag browsing" feature.

srid:
Agreed. What would be the zquery URI equivalent? Here's what I think, at the moment:

* `zquery://tags` -- show a list of tags in the UI, linking to search.html?tag=foo
* `zquery://tags?tree` -- same, but render as a tree
  * Or, alternatively, `zquery://tags?root=/` -- same as before
  * Whereas `zquery://tags?root=math` would render only the math sub-tree

Then have CLI map to those queries.

felko:
Do all the queries need to have a corresponding URI? I would prefer `zquery://tags` to display the tree in all cases, as it is easier to browse.

I like the `zquery://tags?root=math`. I would also allow globbing like `zquery://tags?glob=math/*` since it could happen that you want to display the tree only one level deep.

What would the output look like in the HTML? Will the matched tags be displayed in a tree without any zettels inside? If so, then some kind of `zquery://outline?root=math` might be useful as well. Here is an idea of how that would look like, assuming there are zettels tagged `math/algebra` and `math/calculus`:

```
# math
[.](zquery://search?tag=math)

## algebra
[.](zquery://search?tag=math/algebra)

## calculus
[.](zquery://search?tag=math/calculus)
```

That allows the creation of "dynamic outlines" which are automatically updated as new zettels are created under the `math` tag.

srid:
It is good to have a 1-on-1 map to queries. I think of `neuron query --whatever` simply as an alias for using the full form `neuron query --uri`. Plus, it forces us to think how whatever querying we add to the CLI would impact the HTML view.

Alright, let's do tree by default, and go with `zquery://tags?filter=math` and `zquery://tags?filter=math/*` ("filter" is more appropriate naming here, and consistent with the tag-based filtering of `search?tag=...`).

> What would the output look like in the HTML? 

Tree of links to the tag search page.

> Will the matched tags be displayed in a tree without any zettels inside? 

Why would we show the zettels? The user is querying for tags only, not the zettels.

[...]

PR #121, opened 2020-04-20
Tag tree

felko:
Implements #110.

The query command was implemented like this:
```
neuron query zettel ID
neuron query zettels [TAG]
neuron query tags [TAG] [--tree | --list (default) ]
neuron query --uri URI
```

The `--tree` argument provides a JSON tree where each node contains some information that would be useful for a tag browsing feature in editors, as discussed in the PR.

### DIFF
--- a/neuron.cabal
+++ b/neuron.cabal
@@ -69,6 +69,7 @@ library
   exposed-modules:
     Neuron.CLI
     Neuron.CLI.New
+    Neuron.CLI.Query
     Neuron.CLI.Rib
     Neuron.CLI.Search
     Neuron.CLI.Types
--- a/neuron.cabal
+++ b/neuron.cabal
@@ -62,7 +62,8 @@ common library-common
     megaparsec >= 8.0,
     some,
     dependent-sum,
-    dependent-sum-template
+    dependent-sum-template,
+    parser-combinators
 
 library
   import: library-common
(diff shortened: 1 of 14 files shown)


felko:
Update: The CLI regressed to its previous spec.
Tag queries can only be executed by using `neuron query --uri zquery://tags?...` for now, and the result is a JSON tree in any way.

The JSON tree follows this format:
```
tree ::= {
  "tag": reltag,
  "subTags": [reltag...]
}
reltag ::= {
  "name": string
  "exists": boolean,
  "root": string
}
```

If a zettelkasten contains zettels tagged `science/biology` and `science/math` (but not `science` alone), then the tree for `zquery://tags?filter=science/**` would look like this:
```json
[
  {
    "tag": {
      "exists": false,
      "name": "science"
    },
    "subTags": [
      {
        "exists": true,
        "root": "science",
        "name": "biology"
      },
      {
        "exists": true,
        "root": "science",
        "name": "math"
      }
    ]
  }
]
```
This makes it very easy for editors to implement tag browsing, but the format is a little bit complicated, and contains some duplicate information since the leaves will always have `exists = true` anyway.

felko:
Update: The tree is now under this format:

```
tree ::= {
  "tag": string,
  "count": int,
  "children": [tree...]
}
```
where `count` is the number of zettels associated with each tag.

The view has also been updated to display the count.

Issue #86, opened 2020-04-11
Implement `neuron open <zettelID>`


srid:
Have `neuron open 2011501` open the corresponding locally generated file `2011501.html`. This could be used in editor support.

srid:
Merged in #325

PR #138, opened 2020-04-25
Fix small issues in neuron-search

### LINKED ISSUE #56: `neuron … search` misbehaves when there is only one note
To reproduce:

```
$ mkdir test
$ neuron test new Test
$ neuron test search
```

It will misbehave in a few ways:

* There will be a `[bat error]: 'title': No such file o…` error message
* Attempts to refine the query will generate 0 results
* Selecting the single note will return `./test/title` instead of the path to the note

Once you add more than one note, though, the problem disappears

jared-w:
`neuron rib -wS` spawns a http sever, but this command tried to open a https server. Additionally, the echo command was missing the `$` required to work. Now pressing ctrl-j in `neuron search` works as expected.

While browsing the issues, I also stumbled upon #56 and fixed that as well.

There remains the issue that if there are no markdown files in `NOTESDIR`, the script will exit with an error along the lines of:

```sh
*.md: IO error for operation on *.md: No such file or directory (os error 2)
```

which might not be desirable, but it's certainly a highly unusual circumstance given the nature of `neuron search`. 

Closes #56

### DIFF
--- a/src-bash/neuron-search/default.nix
+++ b/src-bash/neuron-search/default.nix
@@ -14,7 +14,6 @@ pkgs.writeShellScriptBin "neuron-search"
     ${pkgs.ripgrep}/bin/rg --no-heading --no-line-number --with-filename --sort path "''${FILTERBY}" *.md \
       | ${pkgs.fzf}/bin/fzf --tac --no-sort -d ':' -n ''${SEARCHFROMFIELD}.. \
         --preview '${pkgs.bat}/bin/bat --style=plain --color=always {1}' \
-        --bind 'ctrl-j:execute(xdg-open http://localhost:8080/$(echo {1} | ${pkgs.gnused}/bin/sed "s/\.md/.html/g"))+abort' \
       | ${pkgs.gawk}/bin/awk -F: "{printf \"''${NOTESDIR}/%s\", \$1}" \
       | ${pkgs.findutils}/bin/xargs -r ''${OPENCMD}
   ''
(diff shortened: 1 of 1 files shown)


srid:
> Additionally, the echo command was missing the `$` required to work.

Are you using bash? Because your change breaks it for me on [fish](https://github.com/fish-shell/fish-shell).

I think we should just get rid of this hidden ctrl+j feature from `neuron search`. Something like this instead would probably be better: `neuron search --id-only | xargs -n 1 neuron open` (see #86).

jared-w:
> Are you using bash? Because your change breaks it for me on [fish](https://github.com/fish-shell/fish-shell).

Interesting. I think this is because the shellscript doesn't set a shebang and so it uses whatever shell you're using. It would probably work fine and be as useful if an explicit shebang was set to something like bash. This is probably best regardless of what we do with ctrl+j; there are too many shells with subtle (or large) differences to really write a cross-shell script.

Do you want me to add the shebang so it always runs through a consistent interpreter?

> I think we should just get rid of this hidden ctrl+j feature from `neuron search`. Something like this instead would probably be better: `neuron search --id-only | xargs neuron open` (see #86), or even simply `neuron search | xargs xdg-open`.

Does `neuron search -e` not already do this? In fact, now that I think about it, there doesn't seem to be any reason to have the `ctrl+j` feature given the existence of `neuron search -e`.

Should I just remove the `ctrl+j` binding entirely? I am leaning towards this.

srid:
Yea, let's remove it entirely.

jared-w:
Ok. I removed it. Also, since that was the only thing that was causing incompatibility between fish and bash in the script, I opted to not change the `#!${runtimeShell}` shebang behavior of `writeShellScriptBin`. Ironically, this still fixes the original issue, but simply by removing the now unnecessary feature since `neuron search -e` already works. Seems like a win to me.

PR #164, opened 2020-05-08
Prepare everything for a nixpkgs compatible release

maralorn:
This contains the following steps as discussed before:

* Track bash, dhall, javascript and Haskell files which are needed to compile via cabal so that they are included in the hackage package.
* Change neuron-search to a pure bash script that is expected to be in the PATH while compiling the package. And modify the default.nix to respect this.

## To Check
Also your project is kinda elaborate so please look at this carefully because I am not sure that I did everything right.
Some points for you to look at:
* Was I right in including so many packages in the other-modules?
* Is there a better solution then expecting neuron-search to be in PATH at compile time?

## TODO
- [x] Check that this can be built with default.nix
- [x] Check that this can be built just from the sdist file.
- [x] Check that this can be built with a reasonable expression from within nixpkgs. See: https://github.com/NixOS/nixpkgs/pull/87206
- [x] Check that the version string works in nixpkgs as discussed before.


neuron.cabal
```diff
@@ -16,6 +16,11 @@ description:
 extra-source-files:
   README.md
   CHANGELOG.md
+data-files:
+  src-js/search.js
+  src-purescript/hello/index.js
```
srid, line 21: Note that `./src-purescript` is currently unused by the app. (it exists just as a reference implementation to use in future).

srid: Oh nevermind. It *is* used as a demonstration. I'll remove it later.

maralorn:
Okay. I think I am done. I believe if you merge this we can this in nixpkgs within a week.
But of course there is always room for improvement so I’m open for feedback.

srid:
Great! I'm testing your branch. User install works, tested with the following command on a pristine machine

```
 nix-env -iE '_:
  let src = builtins.fetchGit {
    url = "https://github.com/maralorn/neuron"; ref = "fix-nixpkgs"; };
  in import src.outPath { gitRev = src.shortRev; }'
```

It shows the expected version post-install:

```
$ neuron --version
0.3.1.0 (eb6a55f)
```

---

However, development workflow is not functional. In particular, `bin/run --version` (for example), errors out:

```
<interactive>:27:64-67: error:
    Variable not in scope: main :: INTERNAL_GHCID.IO a0
```

I remember seeing this error before ...

srid:
@maralorn Okay, deleting `other-modules` from both the stanzas fixes the INTERNAL_GHCID error for me (`bin/test` also works). 

I guess now I remember why I had left it out. I don't understand exactly why this causes a problem though. Is there a reason (besides suppressing cabal warnings) why we must specify `other-modules` here?

maralorn:
I think it‘s simply that, if you don‘t list them in the cabal file they don‘t get packaged into the sdist. But I may be wrong about that.

maralorn:
Yep, I double checked. Those files have to be listed somewhere as other-modules or exposed-modules to be included in the sdist tarball.
Sadly I don‘t understand what your ghcid setup is doing and why it is failing.

Just one guess: It was my understanding that having the same Source file included into different cabal targets can lead to confusion. As in: Right now the app is not using the library as it is defined in cabal, because it just includes all source files by itself and says they belong to the app. I didn‘t touch that because I didn‘t know how you want it to be, but I think we should try to splite the files between app and library cleanly optimally in src-lib and src-app and then include the library as a dependency of the app. Or we decide to put everything in the lib other then the main in src-bin.

That‘s just a guess and a suggestion. Maybe I understood something wrong.

What do you think?

srid:
> Sadly I don‘t understand what your ghcid setup is doing and why it is failing.

@maralorn Here's some background on why the cabal file is the way it is: https://haskell.zettel.page/2012605.html

In order words, I wanted to use ghcid on both the app and the tests, and have it reload automatically if any of the source files in neuron change. In a normal cabal file, if you run ghcid on the app or the test, it won't reload if the library sources change.

srid:
By the way, I'd like to keep `lib` and `app` (the executable) separate. The idea being that `lib` should be lightweight, and compile on GHCJS, leaving `app` free to do whatever (including depending on pandoc, for instance).

`lib` will be used in other projects; `app` is just the neuron CLI app. Over time I plan to move more modules from `app` to `lib`.

srid:
Actually this may be solvable by fixing the `bin/*` scripts. They currently use `:main`, and replacing that with `Main.main` seems to work (would need a `withArgs` I think) ..

srid:
@maralorn Okay, here's a reasonable workaround: https://github.com/ndmitchell/ghcid/issues/216#issuecomment-447359017

This patch fixes it for me: https://gist.github.com/srid/82ce3e710268279f095873f1e2346b60

maralorn:
Thanks for explaining the thinking behind it. That separation looks very smart to me. I was also already thinking about integrating a bit of neuron into my reflex app.
I believe there are other workarounds to have ghcid reload on multiple targets. I think obelisk does something like that.
But I am totally content with the given fix.

srid:
> I was also already thinking about integrating a bit of neuron into my reflex app.

Interesting. What's your app about? Supporting reflex/ghcjs is why `lib` and `app` are being separated out. 

> I believe there are other workarounds to have ghcid reload on multiple targets. I think obelisk does something like that.

Yea, but I don't think obelisk's `ob run` can be used on this project (aside from creating thunks). I was [told](https://github.com/reflex-frp/reflex-ghci/issues/15#issuecomment-626174959) that multi-package GHC sessions is one of this year's Haskell GSOC projects. Whenever any of these becomes ready, I'd love to switch over.

maralorn:
> Interesting. What's your app about? Supporting reflex/ghcjs is why `lib` and `app` are being separated out.

It‘s a taskwarrior frontend. https://github.com/maralorn/kassandra

PR #166, opened 2020-05-09
MMark -> Pandoc AST

### LINKED ISSUE #137: Switch to Pandoc (commonmark-hs)
Requests for Pandoc arose [a few times](https://github.com/srid/neuron/issues?q=label%3A%22mmark+vs+pandoc%22+).

Let's switch to ~~Pandoc~~ [commonmark-hs](https://github.com/jgm/commonmark-hs) (which Pandoc will eventually be using as its markdown parser).

In order to migrate away from MMark to Pandoc, we will have to rewrite the `replaceLink` extension, which I've refactored to be general and small enough:

https://github.com/srid/neuron/blob/fdf18dd7d6af5ae57546839d7cdeff697a4d12db/src/Text/MMark/Extension/ReplaceLink.hs#L16-L25

Essentially the extension takes a `Map` of links, and for each link it renders the given HTML in the final output (the Map is computed ahead by neuron).

The following should continue to work:

- `z:`, `zquery:`, `<ID>`, etc links
- Markdown YAML metadata (for `title`, `date` and `tags`)

srid:
See how feasible it is to completely switch over from `MMark` to `Pandoc` AST. 

* [x] Markdown to be parsed using commonmark-hs (default); 
  - in future potentially we could support multiple source formats (including Pandoc's native markdown parser, org-mode, reST, etc.)
* [x] Use `HsYaml` to manually parse the Yaml blocks (unless https://github.com/jgm/commonmark-hs/issues/17 gets implemented by someone)
  * [x] Make sure it works with GHCJS
* [x] Profile and identify performance bottlenecks
* [x] Make fontawesome icons work (Pandoc AST cannot represent `<i>` with custom classes).
* [x] Modify the rest of neuron to switch over to the new parser
* [x] Use necessary commonmark extensions (eg: math)
* [x] Port/add unit tests
* [x] Documentation updates

Finally,

* [x] #172 Replace pandoc's writer with reflex-dom-pandoc (to be GHCJS friendly)
  * [x] Footnote links are broken (pandoc writer)

If successful, this will resolve #137.


srid:
HsYAML is rather slow. https://github.com/haskell-hvr/HsYAML/issues/40

image

image

PR #169, opened 2020-05-09
Release 0.4

srid:


maralorn:
Just out of curiosity, what‘s up with the even version numbers? Do you do it like ghc?

srid:
Yup. Shortly, I'll increment the development branch version to 0.5, and later adjust the installation instructions to specify installing from either stable (from nixpkgs) and development versions (from git).

PR #191, opened 2020-05-15
Raw HTML support

srid:
HTML elements must provide an explicit list of attributes (so as to disambiguate from zettel links). Actual work happens at https://github.com/srid/reflex-dom-pandoc/pull/1


srid:
Example of this feature, where we embed an entire Google spreadsheet in a zettel: https://reddit.zettel.page/software-comparison.html#table

srid:
Example of this feature, where we embed an entire Google spreadsheet in a zettel: https://reddit.zettel.page/software-comparison.html#table

PR #216, opened 2020-06-04
Implement a backlinks query for a given Zettel ID

pjones:
Work in progress: do not merge.

Will be useful for editors to show a "nearby zettels" view.  See
felko/neuron-mode#30 for more details.

### DIFF
--- a/neuron/src/app/Neuron/CLI/Types.hs
+++ b/neuron/src/app/Neuron/CLI/Types.hs
@@ -21,6 +21,7 @@ import Data.Default (def)
 import Data.Some
 import Data.TagTree (mkTagPattern)
 import Data.Time
+import qualified Neuron.Zettelkasten.Connection as C
 import Neuron.Zettelkasten.ID (ZettelID, parseZettelID')
 import Neuron.Zettelkasten.ID.Scheme (IDScheme (..))
 import qualified Neuron.Zettelkasten.Query.Error as Q
@@ -134,6 +135,26 @@ commandParser defaultNotesDir today = do
             <|> fmap
               Right
               (fmap (const $ Some $ Q.GraphQuery_Id) $ switch (long "graph" <> help "Get the entire zettelkasten graph as JSON"))
+            <|> fmap
+              Right
+              ( fmap (Some . Q.GraphQuery_BacklinksOf Nothing) $
+                  option
+                    zettelIDReader
+                    ( long "backlinks-of"
+                        <> help "Get backlinks to the given zettel ID"
+                        <> metavar "ID"
+                    )
+              )
+            <|> fmap
+              Right
+              ( fmap (Some . Q.GraphQuery_BacklinksOf (Just C.Folgezettel)) $
+                  option
+                    zettelIDReader
+                    ( long "uplinks-of"
+                        <> help "Get uplinks to the given zettel ID"
+                        <> metavar "ID"
+                    )
+              )
         )
     searchCommand = do
       searchBy <-
--- a/neuron/src/lib/Neuron/Zettelkasten/Graph.hs
+++ b/neuron/src/lib/Neuron/Zettelkasten/Graph.hs
@@ -48,9 +48,14 @@ backlinkForest conn z =
     . G.bfsForestBackwards z
     . G.induceOnEdge (== Just conn)
 
-backlinks :: Connection -> Zettel -> ZettelGraph -> [Zettel]
-backlinks conn z g =
-  G.preSetWithEdgeLabel (Just conn) z g
+backlinks ::
+  (Maybe Connection -> Bool) ->
+  Zettel ->
+  ZettelGraph ->
+  [(Connection, Zettel)]
+backlinks f z g =
+  mapMaybe (\(e, v) -> (,v) <$> e) $
+    G.preSetWithEdgeLabel f z g
 
 -- | Like backlinks but for multiple zettels. More performant than calling
 -- `backlinks` in a loop.
--- a/neuron/src/lib/Neuron/Zettelkasten/Query.hs
+++ b/neuron/src/lib/Neuron/Zettelkasten/Query.hs
@@ -16,6 +16,7 @@ import Data.Aeson
 import qualified Data.Map.Strict as Map
 import Data.TagTree (Tag, tagMatch, tagMatchAny, tagTree)
 import Data.Tree (Tree (..))
+import Neuron.Zettelkasten.Graph (backlinks, getZettel)
 import Neuron.Zettelkasten.Graph.Type
 import Neuron.Zettelkasten.ID
 import Neuron.Zettelkasten.Query.Error (QueryResultError (..))
@@ -46,9 +47,15 @@ runZettelQuery zs = \case
       Map.fromListWith (+) $
         concatMap (\Zettel {..} -> (,1) <$> zettelTags) zs
 
-runGraphQuery :: ZettelGraph -> GraphQuery r -> r
+runGraphQuery :: ZettelGraph -> GraphQuery r -> Either QueryResultError r
 runGraphQuery g = \case
-  GraphQuery_Id -> g
+  GraphQuery_Id -> Right g
+  GraphQuery_BacklinksOf conn zid ->
+    case getZettel zid g of
+      Nothing ->
+        Left $ QueryResultError_NoSuchZettel zid
+      Just z ->
+        Right $ backlinks (maybe isJust (const (== conn)) conn) z g
 
 zettelQueryResultJson ::
   forall r.
@@ -95,19 +102,24 @@ graphQueryResultJson ::
   forall r.
   (ToJSON (GraphQuery r)) =>
   GraphQuery r ->
-  r ->
+  Either QueryResultError r ->
   -- Zettels that cannot be parsed by neuron (and as such are excluded from the graph)
   Map ZettelID ZettelError ->
   Value
-graphQueryResultJson q r skippedZettels =
+graphQueryResultJson q er skippedZettels =
   toJSON $
     object
       [ "query" .= toJSON q,
-        "result" .= resultJson,
+        either
+          (\e -> "error" .= toJSON e)
+          (\r -> "result" .= toJSON (resultJson r))
+          er,
         "skipped" .= skippedZettels
       ]
   where
-    resultJson :: Value
-    resultJson = case q of
+    resultJson :: r -> Value
+    resultJson r = case q of
       GraphQuery_Id ->
+        toJSON r
+      GraphQuery_BacklinksOf _ _ ->
         toJSON r
--- a/neuron/src/lib/Neuron/Zettelkasten/Query/Graph.hs
+++ b/neuron/src/lib/Neuron/Zettelkasten/Query/Graph.hs
@@ -14,13 +14,21 @@ import Data.Aeson.GADT.TH
 import Data.Dependent.Sum.Orphans ()
 import Data.GADT.Compare.TH
 import Data.GADT.Show.TH
+import Neuron.Zettelkasten.Connection
 import Neuron.Zettelkasten.Graph.Type
+import Neuron.Zettelkasten.ID
+import Neuron.Zettelkasten.Zettel
 import Relude
 
 -- | Like `GraphQuery` but focused on the relationship between zettels.
 data GraphQuery r where
   -- | Query the entire graph.
   GraphQuery_Id :: GraphQuery ZettelGraph
+  -- | Query backlinks.
+  GraphQuery_BacklinksOf ::
+    Maybe Connection ->
+    ZettelID ->
+    GraphQuery [(Connection, Zettel)]
 
 deriveJSONGADT ''GraphQuery
 
--- a/neuron/src/lib/Data/Graph/Labelled/Algorithm.hs
+++ b/neuron/src/lib/Data/Graph/Labelled/Algorithm.hs
@@ -50,9 +50,16 @@ preSet (vertexID -> zid) g =
 -- | Return the preset of a vertex, considering only edges with the given label
 --
 -- WARNING: Dont' call this in a loop. For that, use preSetWithEdgeLabelMany
-preSetWithEdgeLabel :: (Eq e, Vertex v, Ord (VertexID v)) => e -> v -> LabelledGraph v e -> [v]
-preSetWithEdgeLabel e v g =
-  preSet v $ induceOnEdge (== e) g
+preSetWithEdgeLabel ::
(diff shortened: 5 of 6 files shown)


neuron/src/lib/Neuron/Zettelkasten/Query/Graph.hs
```diff
@@ -15,12 +15,15 @@ import Data.Dependent.Sum.Orphans ()
...
 import Neuron.Zettelkasten.Graph.Type
+import Neuron.Zettelkasten.ID
 import Relude
 
 -- | Like `GraphQuery` but focused on the relationship between zettels.
 data GraphQuery r where
   -- | Query the entire graph.
   GraphQuery_Id :: GraphQuery ZettelGraph
+  -- | Query backlinks.
+  GraphQuery_BacklinksOf :: ZettelID -> GraphQuery ZettelGraph
```
srid, line 26: This should probably take `Connection` as another argument and return `[Zettel]` instead. You can use the [`backlinks`](https://github.com/srid/neuron/blob/ac46252e0d68a1d7176c223515b04a56f9e5cfa9/neuron/src/lib/Neuron/Zettelkasten/Graph.hs#L48-L52) function to run the query.

srid:
Something to think of: are we looking for non-folgezettel backlinks (the list displayed below the zettel), or also [uplinks](https://neuron.zettel.page/2011503.html) (the tree displayed above the zettel)?

NullSense:
https://github.com/andymatuschak/note-link-janitor Some inspiration.

pjones:
@srid I have the `--backlinks-of` query working.  If you approve of the implementation I will also implement `--folgezettel-of` as well.

What would the query flag be for the `OrdinaryConnection` connection?

neuron/src/lib/Neuron/Zettelkasten/Graph.hs
```diff
@@ -45,11 +45,14 @@ backlinkForest conn z =
...
-backlinks :: Connection -> Zettel -> ZettelGraph -> [Zettel]
+backlinks :: Maybe Connection -> Zettel -> ZettelGraph -> [Zettel]
 backlinks conn z g =
-  filter (not . branches) $ G.preSet z $ G.induceOnEdge (== Just conn) g
+  filter (not . branches) $ G.preSet z $ G.induceOnEdge matcher g
   where
     branches bz = G.hasEdge g z bz
+    matcher = case conn of
+      Nothing -> isJust
+      Just _ -> (== conn)
```
srid, line 55: Rename and replace with:

```haskell
includeConn = maybe isJust (==) conn
```

srid, line 55: Rename and replace with:

```haskell
includeConn = maybe isJust (==) conn
```

neuron/src/lib/Neuron/Zettelkasten/Query.hs
```diff
@@ -49,6 +50,12 @@ runZettelQuery zs = \case
 runGraphQuery :: ZettelGraph -> GraphQuery r -> r
 runGraphQuery g = \case
   GraphQuery_Id -> g
+  GraphQuery_BacklinksOf conn zid ->
+    case getZettel zid g of
+      Nothing -> []
```
srid, line 55: Instead of ignoring the error, this should perhaps raise the error like `runZettelQuery` currently does.

See https://github.com/srid/neuron/blob/4d20756dbb3d66eedc7665b4d37fb668f5d291de/neuron/src/lib/Neuron/Zettelkasten/Query.hs#L30-L34

neuron/src/lib/Neuron/Zettelkasten/Query.hs
```diff
@@ -49,6 +50,12 @@ runZettelQuery zs = \case
...
+      Just z2 ->
+        mapMaybe (\z1 -> (,z1) <$> getConnection z1 z2 g) $
```
srid, line 57: We shouldn't have to lookup the connection a *second* time, if `backlinks` can return it along with the zettel (i.e., `[(Connection, Zettel)]` instead `[Zettel]`).

neuron/src/lib/Neuron/Zettelkasten/Query.hs
```diff
@@ -111,3 +118,5 @@ graphQueryResultJson q r skippedZettels =
     resultJson = case q of
       GraphQuery_Id ->
         toJSON r
+      GraphQuery_BacklinksOf _ _ ->
+        toJSON r
```
srid, line 122: Can you paste the JSON output this generates?

neuron/src/lib/Neuron/Zettelkasten/Query.hs
pjones: @srid Do you have a suggestion for how `backlinks` could return the connection along with the Zettel?  It looks like I'd have to replace `preSet` with something that doesn't discard the graph edges.  I don't see a function that can do what `preSet` does yet preserve the edges.

Would I need to generate a `Forest` and then fold the forest into a list of vertices and edges?

srid: Note that in latest master, backlinks uses `preSetWithEdgeLabelMany` which in turn uses `postSet` (for efficiency; see #245). And `postSet` is defined in the alga library as `Map.findWithDefault Set.empty x . adjacencyMap` - i.e., it just looks up the map (which was produced from the 'inverse' graph). 

You could probably instead do `Map.lookup x . adjacencyMap` which gets you a `Map a e` (i.e., backlinks with connection).

pjones: That sounds good.  I'll rebase and look around.  Thank you.

neuron/src/lib/Neuron/Zettelkasten/Query.hs
pjones:
```
$ ./result/bin/neuron -d guide query --backlinks-of 4a6b25f1
```

```json
{
  "skipped": {},
  "result": [
    [
      "Folgezettel",
      {
        "zettelTags": [
          "walkthrough"
        ],
        "zettelDay": "2020-03-19",
        "zettelID": "2011406",
        "zettelError": {
          "Right": []
        },
        "zettelContent": [],
        "zettelQueries": [
          [
            "ZettelQuery_ZettelByID",
            [
              "2011403",
              "OrdinaryConnection"
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "2012301",
              null
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "2011504",
              "OrdinaryConnection"
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "2013501",
              "OrdinaryConnection"
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "4a6b25f1",
              null
            ]
          ]
        ],
        "zettelTitle": "Creating and Editing zettels",
        "zettelTitleInBody": true
      }
    ],
    [
      "OrdinaryConnection",
      {
        "zettelTags": [],
        "zettelDay": null,
        "zettelID": "6f0f0bcc",
        "zettelError": {
          "Right": []
        },
        "zettelContent": [],
        "zettelQueries": [
          [
            "ZettelQuery_ZettelByID",
            [
              "4a6b25f1",
              "OrdinaryConnection"
            ]
          ]
        ],
        "zettelTitle": "Philosophy",
        "zettelTitleInBody": true
      }
    ]
  ],
  "query": [
    "GraphQuery_BacklinksOf",
    [
      null,
      "4a6b25f1"
    ]
  ]
}
```

```
$ ./result/bin/neuron -d guide query --uplinks-of 4a6b25f1 | jq
```

```json
{
  "skipped": {},
  "result": [
    [
      "Folgezettel",
      {
        "zettelTags": [
          "walkthrough"
        ],
        "zettelDay": "2020-03-19",
        "zettelID": "2011406",
        "zettelError": {
          "Right": []
        },
        "zettelContent": [],
        "zettelQueries": [
          [
            "ZettelQuery_ZettelByID",
            [
              "2011403",
              "OrdinaryConnection"
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "2012301",
              null
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "2011504",
              "OrdinaryConnection"
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "2013501",
              "OrdinaryConnection"
            ]
          ],
          [
            "ZettelQuery_ZettelByID",
            [
              "4a6b25f1",
              null
            ]
          ]
        ],
        "zettelTitle": "Creating and Editing zettels",
        "zettelTitleInBody": true
      }
    ]
  ],
  "query": [
    "GraphQuery_BacklinksOf",
    [
      "Folgezettel",
      "4a6b25f1"
    ]
  ]
}
```

neuron/src/lib/Neuron/Zettelkasten/Graph.hs
pjones: This change has been applied in the latest commit.

neuron/src/lib/Neuron/Zettelkasten/Query.hs
pjones: The `backlinks` function now returns the Zettel *and* connection.

neuron/src/lib/Neuron/Zettelkasten/Query.hs
pjones: The graph query now uses `Either`.

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -134,6 +135,26 @@ commandParser defaultNotesDir today = do
             <|> fmap
               Right
               (fmap (const $ Some $ Q.GraphQuery_Id) $ switch (long "graph" <> help "Get the entire zettelkasten graph as JSON"))
+            <|> fmap
+              Right
+              ( fmap (Some . Q.GraphQuery_BacklinksOf Nothing) $
+                  option
+                    zettelIDReader
+                    ( long "backlinks-of"
+                        <> help "Get a backlink graph of ID"
```
srid, line 144: "Get backlinks to the given zettel ID"

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -134,6 +135,26 @@ commandParser defaultNotesDir today = do
...
+                        <> metavar "ID"
+                    )
+              )
+            <|> fmap
+              Right
+              ( fmap (Some . Q.GraphQuery_BacklinksOf (Just C.Folgezettel)) $
+                  option
+                    zettelIDReader
+                    ( long "uplinks-of"
+                        <> help "Get an uplink graph of ID"
```
srid, line 154: "Get uplinks to the given zettel ID"

neuron/src/lib/Neuron/Zettelkasten/Graph.hs
```diff
@@ -48,9 +49,12 @@ backlinkForest conn z =
...
     . G.induceOnEdge (== Just conn)
 
-backlinks :: Connection -> Zettel -> ZettelGraph -> [Zettel]
+backlinks :: Maybe Connection -> Zettel -> ZettelGraph -> [(Connection, Zettel)]
 backlinks conn z g =
-  G.preSetWithEdgeLabel (Just conn) z g
+  let includeConn = maybe isJust (const (== conn)) conn
+      g' = LAM.transpose $ G.getGraph $ G.induceOnEdge includeConn g
+      ns = Map.toList $ Map.findWithDefault mempty (G.vertexID z) $ LAM.adjacencyMap g'
+   in mapMaybe (\(v, e) -> (,) <$> e <*> G.findVertex v g) ns
```
srid, line 57: This implementation should be moved to `Data.Graph.Labelled.Algorithm`. You can just put them in the existing `preSetWithEdgeLabel` function. 

Also, you should use `getVertex` here instead of `findVertex`. The former works with the guarantee that the vertex exists -- which is the case here -- and doesn't have to deal with Maybe. This way you also don't have to ignore the error via the Applicative lift.

Aside: Maybe we should just pass the filter function directly -- i.e., `(Maybe Connection -> Bool)` instead `Maybe Connection` -- but it is okay if that's not done in this PR>

srid: > maybe isJust (const (== conn)) conn

Isn't this the same as `maybe isJust (==) conn`?

neuron/src/app/Neuron/CLI/Types.hs
pjones: Updated wording in the latest commit.

neuron/src/app/Neuron/CLI/Types.hs
pjones: Updated wording in the latest commit.

neuron/src/lib/Neuron/Zettelkasten/Graph.hs
pjones: I can move about half of the implementation into `Data.Graph.Labelled.Algorithm` due to `preSetWithEdgeLabel` being polymorphic in the edge type.

In `backlinks` the edge type is specialized to `Maybe Connection` which is why I post process the `postSet` by removing `Maybe` from the edges.  The other issue with moving more of the code into `preSetWithEdgeLabel` is that `LAM.traspose` imposes more constraints on the edge type than currently exist.  For example, `e` would need to be an instance of `Monoid` and `Ord`.

How does that sound to you?

srid: `preSetWithEdgeLabel` can just take those extra constraints on `e`. If you pass `(e -> Bool)` instead of `Maybe Connection` (as the filter) to it, the function can remain polymorphic - and you can then move all of the implementation.

pjones: Most of the logic has been moved into `preSetWithEdgeLabel`.  Please let me know what you think.

srid: @pjones Almost there. The `backlinks` function should take `(Maybe Connection -> Bool)`  instead of `Maybe Connection` - just to avoid potential confusion.

(`Maybe Connection` is used as the edge type only because the alga type requires it to be a monoid - where the empty value signifies "no connection"; using the same type for the filter can be confusing).

pjones: Good idea.  The latest commit has this change.

srid: Looks good to me. Once you mark it non-WIP, I'll merge it.

pjones: Done.

PR #221, opened 2020-06-09
Render raw zettel on parse errors

### LINKED ISSUE #215: Generate raw zettel if it fails to parse
Currently if a zettel fails to parse (invalid markdown, or bad yaml), we skip HTML generation of it, and report it in the z-index. This is problematic in two ways:

* All zettels linking to this failing zettel will now show up in the z-index error report because they now link to a "non-existing" zettel (though this is in fact false; the zettel only fails to parse, but it actually exists).
* Its old HTML would still be lying around. And if the zettelkasten is publicly hosted, the outdated version would continue to be accessible in search engine index.

One solution is to generate the HTML anyway, but put the whole (unparsed) markdown _text_ inside a `<pre>` block, along with rendering the parse error above it.

srid:
Resolves #215

- [x] Links to raw zettels (which use a placeholder title) should render with a red border
- [x] Eliminate dead code and refactor

### DIFF
--- a/neuron/src/app/Main.hs
+++ b/neuron/src/app/Main.hs
@@ -33,14 +33,14 @@ generateMainSite :: Action ()
 generateMainSite = do
   Rib.buildStaticFiles ["static/**"]
   config <- Config.getConfig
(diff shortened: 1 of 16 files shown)


srid:
image

PR #227, opened 2020-06-11
added support for $EDITOR not being an executable

TheAceShinigami:
it's kind of a hack, but in order to well support whatever they have as their $EDITOR you'll need to interpret shell faithfully, so it has to be a bit hacky. I wanted to use `sh`, but sh doesn't have the `-c` option (I don't think it's posix). I suppose this could be improved by checking the shell environment variable and acting accordingly, but supporting shells that don't have the "-c" option seems impossible. I figured trying to support all shells was futile so I just passed it to bash because that covers 99.9% of cases. (if your $EDITOR is not bash compliant I'm curious what black magic you are doing)


neuron/src/app/Neuron/CLI/New.hs
```diff
@@ -65,12 +65,12 @@ newZettelFile NewCommand {..} = do
...
-        Just editorExe ->
-          pure $ editAction editorExe
-    editAction editorExe path = do
+        Just editor ->
+          pure $ editAction editor
+    editAction editor path = do
       -- Show it first in case the editor launch fails
       showAction path
-      executeFile editorExe True [path] Nothing
+      executeFile "bash" True ["-c", editor ++ ' ' : path ] Nothing
```
srid, line 73: Wouldn't this break if `path` had whitespace in it?

TheAceShinigami: does path not create escape characters? How was it working before?

neuron/src/app/Neuron/CLI/New.hs
```diff
@@ -45,11 +44,11 @@ newZettelFile NewCommand {..} = do
   case mzid of
     Left e -> die $ show e
     Right zid -> do
-      path <- zettelPath zid
+      (notesDir, filename) <- zettelPath zid
```
srid, line 47: Let's get rid of the `zettelPath` function, and inline its body here.

Issue #189, opened 2020-05-15
Inline tags


srid:
Allow using tags inline in the zettel content.

Example:

> Gaslighting is enabled by the victim's **@psychology/cognitive-distortion**
without which the victimizer is rendered powerless to influence their victim.

Meta tags should be supported, but now we parse for `@<tag>` in the body as well.

geier:
I'd like to put `#<tag>` forward as an alternative option. Not only have twitter and similar service made the `#<tag>` syntax ubiquitous, but also other note taking apps such as [vscode-markdown-notes](https://github.com/kortina/vscode-markdown-notes#find-all-references-to-tag) (used by foam), [Zettlr](https://www.zettlr.com/features), [Obsidian](https://obsidian.md/features),  [Roam Research](https://roamresearch.com/) and probably many more use it.

srid:
Note to self: internally, tags should be represented as yet another query -- no different from regular link queries. Just as links establish edges on the graph, so too will tags introduce a tag-labelled edge.

For example, the inline text `#daily` can be transformed internally into `<z:tag/daily>`, and then neuron can process tags in the same manner as links. Incidentally this makes implementing inline tags simpler.

It was @jared-w who prompted me into thinking about this, and maybe he has further ideas.

PR #230, opened 2020-06-11
Make title and YAML block optional

### LINKED ISSUE #181: Make title optional
In the absence of the title, use the first few words of the first paragraph. Consequently, the entire metadata block becomes optional. And people can just start typing notes with nothing but their thoughts.

cf. https://lobste.rs/s/kydg6q/neuron_0_4_zettelkasten_note_management

> Titles for cards. These, to me, are an error and lead to ideas being discarded that have value in context. Often it’s much easier to have an idea in context and to write it down. Forcing titling of ideas leads to ideas being discarded that would otherwise (in context) be valuable.

srid:
If no title is specified, use the H1 title (as raw string). If that is missing, use the first few letters of the first paragraph. Otherwise, use "Untitled".

Resolves #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Introduce new "uplink tree" view (#195)
 - Resilient error handling (#202, #215)
 - Added `neuron query --graph` to get the entire graph as JSON export
-- YAML block is now optional, and native Markdown H1 titles are supported (#230)
+- YAML block is now optional; title is also optional, while native Markdown H1 titles are now supported (#230)
 - Bug fixes
   - Fix 'neuron new' generating invalid Markdown when title contains special characters (#163)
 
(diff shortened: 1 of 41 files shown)


dpaetzel:
Hi, let me start by saying that I just recently discovered `neuron` and I'm a big fan :slightly_smiling_face: Great work!

A question though: What is the reasoning behind not putting titles specified via the CLI argument into the YAML any more? (I'm just curious :wink: )

srid:
@dpaetzel The thinking is that going forward we would use the new-style title (`# The title`) *by default*. In conjunction with tags also moving to the body (see #189), the YAML would become strictly unnecessary (except for specifying date).

If that's not desirable for some users' workflow, then we could certainly add an option to `neuron new` like `--title-in-yaml` that preserves the old behaviour (PR welcome!).

dpaetzel:
> the YAML would become strictly unnecessary (except for specifying date)

Makes a lot of sense to me! :)

PR #249, opened 2020-06-25
Document systemd+home-manager usage

Nadrieril:


srid:
Nice. Can you put this on its own zettel?

(In future, I plan to create a 'Deployment' zettel, and then your zettel could be put under that along with CI/CD setup zettels)

Nadrieril:
If I do, from where should I link to it?

srid:
For now, link to it from "Declarative Install".

>

Nadrieril:
Done

PR #258, opened 2020-06-29
Make graph queries output JSON that is consistent with zettel queries

felko:
Fixes #254, needed for <https://github.com/felko/neuron-mode/issues/30>

The new JSON output looks like this:

- `--backlinks-of`
```json
{
  "skipped": {},
  "result": [
    {
      "path": "guide/2011402.md",
      "connection": "Folgezettel",
      "day": "2020-03-19",
      "id": "2011402",
      "title": "Guide",
      "tags": []
    },
    {
      "path": "guide/2011406.md",
      "connection": "OrdinaryConnection",
      "day": "2020-03-19",
      "id": "2011406",
      "title": "Creating and Editing zettels",
      "tags": [
        "walkthrough"
      ]
    },
    ...
  ],
  "query": [
    "GraphQuery_BacklinksOf",
    [
      null,
      "2011403"
    ]
  ]
}
```
- `--uplinks-of`
```json
{
  "skipped": {},
  "result": [
    {
      "path": "guide/2011402.md",
      "connection": "Folgezettel",
      "day": "2020-03-19",
      "id": "2011402",
      "title": "Guide",
      "tags": []
    }
  ],
  "query": [
    "GraphQuery_BacklinksOf",
    [
      "Folgezettel",
      "2011403"
    ]
  ]
}
```

- `--graph`
```json
{
  "skipped": {},
  "result": {
    "2011402": {
      "2013501": "Folgezettel",
      "2011404": "Folgezettel",
      "2011403": "Folgezettel",
      "cc1f7ecf": "Folgezettel",
      "2011405": "Folgezettel",
      "2011406": "Folgezettel",
      "2017401": "OrdinaryConnection"
    },
    "2013501": {},
    "6f0f0bcc": {
      "4a6b25f1": "OrdinaryConnection",
      "041726b3": "OrdinaryConnection"
    },
    ...
  },
  "query": [
    "GraphQuery_Id",
    []
  ]
}
```

Note that for the graph, the information about the zettels themselves do not appear anymore, so we only get their IDs (it only outputs the adjacency map). Not sure whether this is a good decision or not, you tell me.

PR #265, opened 2020-07-01
Normalize JSON output

srid:
Use `ToJSON` everywhere.

* [X] Revert back to initial graph JSON output (address [this regression](https://github.com/srid/neuron/pull/258#issuecomment-652418288))
* [x] Normalize `query --backlinks-of 67c6f7a0`
* [x] Normalize `query --uri "z:zettels?tag=walkthrough"`


srid:
What the JSON looks like when using Aeson's generic deriving:

image

image

image

Issue #262, opened 2020-06-30
Switch from reflex-platform to direct nixpkgs


srid:
Go back to how it was pre-reflex-platform days, but use reflex-dom-core, reflex, and friends from nixpkgs (instead of reflex-platform). 

Replace project.nix & ./dep/* with a new default.nix.

The following should continue to work:

- [x] `bin/run` & `bin/test` scripts should continue to reload when any of the source files change
- [x] `nix-build` when run a second time should not build anything new
- [x] neuron [install command](https://neuron.zettel.page/2011501.html) should not regress to installing more than needed ([see here](https://github.com/srid/neuron/blob/4b29d4081daa0d460a540e5e4846cd52ce9ddb83/default.nix#L6-L30))
  - On a fresh VM, installing neuron produces a /nix that weighs ~1.9G.
- [x] ci.nix should handle both `nix-build` and `nix-shell` cases for cache
- [x] Adjust "Developing on reflex-dom-pandoc" section of README accordingly

/cc @maralorn

srid:
This is done #280, so I'll close.

@maralorn AFAIU, only rib and commonmark needs to be released to hackage before neuron can be updated on nixpkgs, right?

maralorn:
@srid, and reflex-dom-pandoc, if it isn't already.

PR #280, opened 2020-07-11
Disable use of reflex-platform

maralorn:
This change migrates the default.nix back to a configuration that does
not use reflex-platform.

The default.nix know tries to be similar to how neuron would be build in
nixpkgs.

The pinned nixpkgs version is now pinned to a very recent version of
nixos unstable.

The default ghc on nixpkgs right now is 8.8.3, so this change also
includes some changes to make compilation possible with that ghc and the
newest package versions from nixpkgs.

---

This addresses #262. 

## TODO

- [x] Rebase on latest master
- [x] Address all requirements from #262
  - [x] bin/run & bin/test scripts should continue to reload when any of the source files change
  - [x] nix-build when run a second time should not build anything new
  - [x] neuron install command should not regress to installing more than needed (see here)
       On a fresh VM, installing neuron produces a /nix that weighs ~1.9G.
  - [x] ci.nix should handle both nix-build and nix-shell cases for cache
  - [x] Adjust "Developing on reflex-dom-pandoc" section of README accordingly
- [x] Remove code and files not needed anymore because of this change


maralorn:
Well, this is a bad situation.

I am 95% sure, that the nix expressions are correct (in some fuzzy sense). The ci.nix builds without problems on my machine.

The problem is, that there is only one reasonable commit of nixpkgs to use, which already includes the newest pandoc-types. Going back to older nixpkgs commits would require more temporary fixes that seem pretty annoying to me.

But this special commit has apparently completely unrelated test issues on hydra.

srid:
@maralorn There is no reason to remain in an older nixpkgs. I generally always use the latest nixpkgs commit ([from here](https://status.nixos.org/)) if I can. Otherwise, I don't understand what you mean by "bad situation".

project.nix
```diff
@@ -1,211 +1,120 @@
...
-    # put the latest commit sha of gitignore Nix library here:
-    rev = "00b237f";
-    # use what nix suggests in the mismatch message here:
-    sha256 = "sha256:186pvp1y5fid8mm8c7ycjzwzhv7i6s3hh33rbi05ggrs7r3as3yy";
+let nixpkgsRev = "dc80d7b";
+in {
+  pkgs ? import (builtins.fetchTarball "https://github.com/nixos/nixpkgs/archive/${nixpkgsRev}.tar.gz") { },
+  # Cabal project name
+  name ? "neuron",
+  compiler ? pkgs.haskell.packages.ghc883,
```
srid, line 6: Any reason why we are not using the default compiler (`pkgs.haskellPackages`)? Because then you can take advantage of nixos cache for Haskell packages.

maralorn:
@srid, well the bad situation is, that I have finished the PR from my point of view. And would like it be merged, but now the tests are failing. Especially since I actually believe that the test error is not deterministic.

default.nix
```diff
@@ -1,32 +1 @@
...
-  neuron = ghc.neuron;
-in 
-  neuron.overrideDerivation (drv: {
-    # Avoid transitive runtime dependency on the whole GHC distribution due to
-    # Cabal's `Path_*` module thingy. For details, see:
-    # https://github.com/NixOS/nixpkgs/blob/46405e7952c4b41ca0ba9c670fe9a84e8a5b3554/pkgs/development/tools/pandoc/default.nix#L13-L28
-    #
-    # In order to keep this list up to date, use nix-store and why-depends as
-    # explained here: https://www.srid.ca/04b88e01.html
-    disallowedReferences = [
```
srid, line 13: These comments got lost in the diff.

project.nix
```diff
@@ -1,211 +1,120 @@
...
-    regex-pcre-builtin = hackGet ./dep/regex-pcre-builtin;
-
-    algebraic-graphs = hackGet ./dep/alga;
-    clay = hackGet ./dep/clay;
-    HsYAML = hackGet ./dep/HsYAML;
-
-    rib = hackGet ./dep/rib;
+  overrides = self: super: {
+    # We include rib because it is quite tightly coupled with neuron development
+    rib = doJailbreak (self.callCabal2nix "rib" sources.rib { });
```
srid, line 44: Why do we need to jailbreak `rib`?

project.nix
maralorn: It is the default compiler. I just used that because that's what I found. I can gladly replace it with pkgs.haskellPackages, though I don‘t expect that will change anything regarding the caching.

srid: Yes, let's just use `pkgs.haskellPackages`.

project.nix
maralorn: rib pins `dhall 1.30.*` and `relude 0.6.*` apparently it builds with the newer nixpkgs. So I guess we should loosen the constraints in rib and bump the pin?

srid: Yup, definitely favour loosening the constraints on rib and reflex-dom-pandoc.

srid:
CI passes.

project.nix
srid: @maralorn This was already done? https://github.com/srid/rib/commit/42bc7c3e6b092cb1640b078b772ac09720cf8825

You can update the rib thunk using obelisk.

maralorn: Yeah, cool.

maralorn: What do you think about tracking nixpkgs via an ob thunk as well?

PR #288, opened 2020-07-15
Fix evaluation in pure evaluation mode

davidtwco:
Recent changes since #264 have broken evaluation in pure evaluation mode (required when using this project from a Nix flake), this PR fixes that regression. If maintaining the `sha256` argument to `fetchTarball` is bothersome, then you might consider adopting [`niv`](https://github.com/nmattia/niv) to make updating the pinned external repositories easier.

### DIFF
--- a/project.nix
+++ b/project.nix
@@ -1,6 +1,14 @@
-let nixpkgsRev = "8d05772";
+let 
+  nixpkgsSrc = builtins.fetchTarball {
+    url = "https://github.com/nixos/nixpkgs/archive/8d05772.tar.gz";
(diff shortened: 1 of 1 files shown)


davidtwco:
> Hey, thanks! By the way, how are you installing neuron with flakes? Do you have a link to a repo?

I installed Neuron [in this commit](https://github.com/davidtwco/veritas/commit/48c7cd2a145debb8bc11d2976dbbc978a506ce76) (since changed to this repository rather then my branch for my previous PR; and the lockfile updated to a more recent commit). 

In this case, without this repository having a flake (which might not make sense until flakes are no longer experimental), my use of Neuton through flakes amounts to having Nix manage the pinned commit and evaluate the expressions with some greater restrictions.

PR #293, opened 2020-07-17
Setup automated nightly releases

### LINKED ISSUE #292: Add binary artifacts to releases
I would like download linux bundles of `neuron` in CI outside of Github (https://funprog.zulipchat.com/#narrow/stream/231929-Neuron/topic/Gitlab.20Pages/near/204132793), but Github API seems to require authentification for that - would it be possible to add artifacts from neuron's CI into (publicly available) releases?

TheMatten:
Fixes #292.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,7 @@ jobs:
   create-nightly-release:
     runs-on: ubuntu-latest
     needs: build
-    if: ${{ github.ref == 'refs/heads/master' && runner.os == 'Linux' }}
+    if: ${{ github.ref == 'refs/heads/master' }}
     steps:
     - name: Download latest linux bundle artifact
       uses: actions/download-artifact@v2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
       with:
         name: neuron-bundle-linux
         path: ~/bundle/neuron
-  create-nightly-release:
+  update-nightly-release:
     runs-on: ubuntu-latest
     needs: build
     if: ${{ github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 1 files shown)


srid:
@TheMatten Does this upload to the same "nightly" release on Github, or does it create a new release for every master commit? Can't tell easily by looking at the diff. 

Do you have URLs to the CI action that fully ran this? Otherwise, I'll have to test it in a separate branch.

TheMatten:
I'm using approach from https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts to get artifact from previous job for releasing. When it comes to releases themselves,
```yaml
    - name: Create new nightly release
      id: release
      uses: actions/create-release@v1
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      with:
        tag: nightly-${{ steps.date.outputs.date }}
        release-name: nightly-${{ steps.date.outputs.date }}
        prerelease: true
```
should create new release named `nigthly-012345678901`, where `012345678901` is date and time of build. That job is run on every push and PR as specified on top of workflow, so it should be only once per push to `master` if I understand it right.

I haven't really tested this myself, because it seems like I would have to setup cachix secrets myself - testing it on separate branch sounds like the easiest option.

srid:
Okay, I'll test it on a separate branch.

> should create new release named nigthly-012345678901, where 012345678901 is date and time of build

Hmm, that concerns me. The master branch currentl has 871 commits. If he had this setup from the beginning, wouldn't this have created 871 releases on GitHub? Not sure if that's desirable exactly. I was thinking more along the lines of having *one* release called "nightly", and having the CI job *overwrite* it (it could put date as a file in the zip file for reference).

TheMatten:
Seems like https://github.com/meeDamian/github-release could do that - to keep link to latest bundle stable, we could change release description to contain current build date, instead of changing name of bundle itself.

srid:
I'm testing this branch with a few more changes of mine, and so far it doesn't seem like to this action is *overwriting* an existing release. It seems to create a new release (albeit with same title) on every action run. 

image

srid:
Merged in #295

PR #295, opened 2020-07-17
Setup automated nightly releases (tested)

srid:
Testing #293 /cc @TheMatten


srid:
> tar: ~/bundle/neuron: No such file or directory

No idea what's going on here. The previous step using `actions/upload-artifact@v2` has no trouble using this path, but the subsequent step using `meeDamian/github-release@v2.0.3` fails.

srid:
Okay, success. https://github.com/srid/neuron/releases/tag/nightly

Issue #146, opened 2020-04-27
Publish docker image in CI


srid:
CI should automatically publish the docker image from the project root's `docker.nix`. 

In effect, run these commands in CI:

```
docker load -i $(nix-build docker.nix)
docker login -u sridca -p ${DOCKER_PASS}
docker push sridca/neuron:latest
```

srid:
Running neuron through docker is as simple as:

```
docker run --rm -i -p 8080:8080 -v ~/zettelkasten:/zettelkasten sridca/neuron:test neuron rib -ws 0.0.0.0:8080
```

It will download only about 80MB of image.

TheMatten:
Docker seems to be [preinstalled](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md) on Ubuntu runners, so adding step
```yaml
- name: Publish nightly Docker image
  run: |
    docker load -i $(nix-build docker.nix --arg tag '"nightly"')
    docker login -u sridca -p ${DOCKER_PASS}
    docker push sridca/neuron:nightly
```
should be enough?

srid:
Done in `master`.  Image name: `sridca/neuron` will auto-update in CI.

```
$ sudo docker run -i sridca/neuron neuron --version
0.5.11.1
```

@TheMatten You can use this for GitLab CI.

I'll close this issue once documentation is updated (alternative install; and describe docker-based usage).

TheMatten:
Done - and it works :tada: : https://gitlab.com/thematten/neuron-template/-/jobs/646875219

TheMatten:
Should I update #298 or are those docs changes meant to be made separately?

srid:
I'll write the docs, as well as make some changes to #298.

PR #298, opened 2020-07-20
Update "Automatic Publishing" to include GL Pages

TheMatten:
Links to [modified version of `neuron-template`](https://gitlab.com/thematten/neuron-template) that works with GitLab Pages.


srid:
I think #146 should be done before we merge this. Otherwise, gitlab users may not always be using the latest nightly image (I'm assuming that the [``thematten/neuron-shell`](https://gitlab.com/thematten/neuron-template/-/blob/master/.gitlab-ci.yml) image is built and pushed manually).

TheMatten:
:100: - CI here could be set up to build new images, passing in version as `tag`.

PR #299, opened 2020-07-20
Make `docker.nix` configurable, add option to include shell

TheMatten:
This makes it more convenient to build custom images (possibly with shell included) using `--arg` flags in `nix-build`.
- [x] investigate possibility of lowering shell footprint

### DIFF
--- a/docker.nix
+++ b/docker.nix
@@ -1,17 +1,18 @@
 # Builds a docker image containing the neuron executable
 # Run as:
-#   docker load -i $(nix-build docker.nix)
+#   docker load -i $(
+#     nix-build docker.nix \
+#       --arg name '"<image name>"' \
+#       --arg tag '"<image tag>"'
+#   )
 let
   pkgs = import <nixpkgs> {};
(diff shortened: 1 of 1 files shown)


srid:
For what usecase are you using `includeShell` for? How big does the image get when includeShell is turned on?

TheMatten:
Image is 94.6 MB, compared to 80.75 MB of original - not that much, but it's dead weight for someone not interested in `bash`. `docker image` shows (uncompressed?) sizes 406 MB and 366 MB.

srid:
Okay. What's the use case for `includeShell` though? Is it required for Gitlab CI?

TheMatten:
Ah, sorry - yeah, it tries to run some setup with it.

srid:
Okay, let's get rid of the `includeShell` flag, and always include `pkgs.coreutils pkgs.bash_5` in the image, adding a comment above them explaining why we include these packages by default. The size different isn't much, so it should be fine.

PR #304, opened 2020-07-24
Update commonmark deps to latest version

maralorn:


srid:
@maralorn Can you increment the version (0.6.0.1) and update CHANGELOG?

neuron/neuron.cabal
```diff
@@ -48,8 +48,8 @@ common library-common
     filepattern,
     megaparsec,
     commonmark ^>= 0.1.0.0,
```
srid, line 50: Shouldn't `commonmark` have a version constraint similar to `commonmark-extensions` and `commonmark-pandoc`?

maralorn: Apparently jgm is not keeping the version numbers in sync.

PR #317, opened 2020-08-02
Open locally generated route

amir:
Addressing #86

### DIFF
--- a/neuron/src/app/Neuron/CLI/Open.hs
+++ b/neuron/src/app/Neuron/CLI/Open.hs
@@ -19,7 +19,7 @@ import System.Posix.Process
 
 openLocallyGeneratedFile :: OpenCommand -> Action ()
 openLocallyGeneratedFile OpenCommand {..} = do
-  let htmlFile = withSome route $ \r -> T.unpack $ routeUrlRel r
+  let htmlFile = T.unpack $ routeUrlRel `foldSome` route
   indexHtmlPath <- fmap (</> htmlFile) ribOutputDir
   let opener = if os == "darwin" then "open" else "xdg-open"
   liftIO $ executeFile opener True [indexHtmlPath] Nothing
--- a/neuron/src/app/Neuron/CLI/Open.hs
+++ b/neuron/src/app/Neuron/CLI/Open.hs
@@ -1,0 +1,23 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RecordWildCards   #-}
+
+module Neuron.CLI.Open
+  ( openLocallyGeneratedFile,
+  )
+where
+
+import qualified Data.Text as T
+import Development.Shake (Action)
+import Neuron.CLI.Types (OpenCommand (..))
+import Relude
+import Rib.Shake (ribOutputDir)
+import System.FilePath
+import System.Info (os)
+import System.Posix.Process
+
+openLocallyGeneratedFile :: OpenCommand -> Action ()
+openLocallyGeneratedFile OpenCommand {..} = do
+  let htmlId = maybe "index" T.unpack zettelId
+  indexHtmlPath <- fmap (</> (htmlId ++ ".html")) ribOutputDir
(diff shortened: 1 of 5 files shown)


srid:
@amir Could you update the ChangeLog?

neuron/src/app/Neuron/CLI/Open.hs
```diff
@@ -0,0 +1,23 @@
...
+import Neuron.CLI.Types (OpenCommand (..))
+import Relude
+import Rib.Shake (ribOutputDir)
+import System.FilePath
+import System.Info (os)
+import System.Posix.Process
+
+openLocallyGeneratedFile :: OpenCommand -> Action ()
+openLocallyGeneratedFile OpenCommand {..} = do
+  let htmlId = maybe "index" T.unpack zettelId
```
srid, line 20: This seems to break `neuron open`. It should open `z-index.html`. But this PR makes it open `index.html` (which may not even exist).

Also you should construct a route, and then use `routeUrl`.

Do something like this:

```haskell
htmlFile = Rib.routeUrl $ maybe Route_ZIndex Route_Zettel zettelId
```

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -60,11 +61,16 @@ data SearchBy
   | SearchByContent
   deriving (Eq, Show)
 
+data OpenCommand = OpenCommand
+  { zettelId :: Maybe Text
```
srid, line 65: This should be `Maybe ZettelID`.

srid: Or even just `Route`. Which would allow you to open other pages, like the search page.

srid: We might have to change the CLI spec:

```
# opens zindex
neuron open
# opens a particular zettel
neuron open --id foobar
# opens the search page
neuron open --search
```

neuron/src/app/Neuron/CLI/Open.hs
amir: Opening `index.html` in case of no arguments is to preserve backward compatibility as this is what's happening right now: https://github.com/srid/neuron/blob/2d7c43510021d3cdeda8aed3468ba17a368b5140/neuron/src/app/Neuron/CLI.hs#L65 it's not opening `z-index.html`.

srid: Ah, I see. That's buggy code in `master`. If index.md doesn't exist (not all zettelkastens have them), `neuron open` will break. We should just make it open the z-index (which always exists) by default. Users can do `neuron open --id index` to open the index zettel explicitly.

srid: Alternatively, we could make it open index.html *if* the index zettel exists (open zindex otherwise). In which case, we will have to add a `--zindex` option to the CLI spec below, as a way to explicitly open the z-index even if index.md exists.

srid: You can get the index zettel like this: https://github.com/srid/neuron/blob/cf1dfa8b28ad6e9d011d46015bfd80f1fb6c2552/neuron/src/app/Neuron/Web/View.hs#L103

neuron/src/app/Neuron/CLI/Open.hs
```diff
@@ -0,0 +1,25 @@
...
+import Relude
+import Rib.Route (routeUrlRel)
+import Rib.Shake (ribOutputDir)
+import System.FilePath
+import System.Info (os)
+import System.Posix.Process
+
+openLocallyGeneratedFile :: OpenCommand -> Action ()
+openLocallyGeneratedFile OpenCommand {..} = do
+  let htmlFile = withSome route $ \r -> T.unpack $ routeUrlRel r
```
srid, line 22:
```haskell
let htmlFile = T.unpack . routeUrlRel `foldSome` route
```

srid:
Merged in #325

Issue #158, opened 2020-05-02
Add unlisted property to zettels


srid:
Add an `unlisted` metadata property to effectivley hide the zettel from the z-index, backlinks, etc.

```
---
title: My blog post WIP
unlisted: true
--
...
```

This is mainly to allow writing of draft zettels (such as blog posts yet to be published).

To implement, basically exclude such zettels from the graph, but generate their HTML as usual.

srid:
Merged in #325

PR #318, opened 2020-08-02
Add an unlisted metadata property

amir:
Addresses #158

### DIFF
--- a/neuron/src/app/Neuron/Reader/Org.hs
+++ b/neuron/src/app/Neuron/Reader/Org.hs
@@ -15,12 +15,12 @@ where
 import qualified Data.Map as Map
 import Data.TagTree (Tag (Tag))
 import Data.Tagged
+import Data.Text (toLower)
 import Data.Time.Calendar (Day)
 import Neuron.Reader.Type (ZettelParseError, ZettelReader)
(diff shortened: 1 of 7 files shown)


srid:
@amir Does this still generate the .html for the unlisted zettel?

amir:
@srid yes, it does.

neuron/src/app/Neuron/Reader/Org.hs
```diff
@@ -39,8 +40,12 @@ extractMetadata doc
...
         tags = fmap Tag . words <$> lookup "tags" properties
+        unlisted = parseUnlisted <$> lookup "unlisted" properties
     pure $ Just Meta {..}
   | otherwise = pure Nothing
   where
     parseDate :: Text -> Either ZettelParseError Day
     parseDate date = maybeToRight (Tagged $ "Invalid date format: " <> date) $ parseZettelDate @Maybe date
+
+    parseUnlisted :: Text -> Bool
+    parseUnlisted a = toLower a == "true"
```
srid, line 51: @felko Does this look right? I'm not familiar with org.

srid:
Merged in #325

PR #331, opened 2020-08-11
Add "created" metadata key, replacing "date", for full date-time support

### LINKED ISSUE #257: Extend Metadata 'date' to include hours/minute/seconds ?
There are probably good reasons, to just have the day as a reference. But maybe there are scenarios, where the time-of-day also is useful. For example, just for visualizing the 'timeline' of creating new zettels on one specific day (?)

Also, maybe there should be some documentation, why _not_ modifying the date in metadata is an important concept (if it is..).
I recognize, that extending date to include the time also needs some effort for the timeline-queries, but maybe these could initially ignore time (?)

A zettelkasten user is of course free, to edit 'date' at any moment, and a user can also add dates in the body of the zettel-text (by convention)... I don't know yet what the zettelkasten philosophy is saying about these issues.

clojj:
"legacy" tags 'date' still supported
if 'created' is manually edited to yyyy-dd-mm (no time), it is parsed as well

Fixes #257

### DIFF
--- a/neuron/src/lib/Neuron/Zettelkasten/Query/Parser.hs
+++ b/neuron/src/lib/Neuron/Zettelkasten/Query/Parser.hs
@@ -45,7 +45,7 @@ queryFromURILink (URILink linkText uri) =
       let mconn = if proto == "zcf" then Just OrdinaryConnection else Nothing
       pure $ Just $ Some $ ZettelQuery_ZettelByID zid mconn
     -- Legacy links
-    Just proto | proto `elem` ["zquery", "zcfquery"] ->
+    Just proto | not isAutoLink && proto `elem` ["zquery", "zcfquery"] ->
       case uriHost uri of
         Right "search" -> do
           let mconn = if proto == "zcfquery" then Just OrdinaryConnection else Nothing


clojj:
again, great dev experience so far !

guide/2011505.md
```diff
@@ -18,18 +18,20 @@ Tags can be also be nested; see <535407ad>.
...
 ---
-date: 2020-04-07
+created: 2020-04-07T21:42
 tags:
   - journal
 ---
 ```
 
+Zettel with 'date' tags are still supported, that is they will be rendered correctly.
+
```
srid, line 34: This should be in a footnote.

neuron/src/app/Neuron/CLI/New.hs
```diff
@@ -100,11 +100,11 @@ defaultZettelContent format day mtitle = case format of
...
         "    :PROPERTIES:",
-        "    :Date: " <> date,
+        "    :Created: " <> createdAt,
         "    :END:",
         "\n"
       ]
   where
-    date = formatZettelDate day
-    defaultTitleName = "Zettel created on " <> date
+    createdAt = formatZettelLocalTime created
```
srid, line 108: rename `createdAt` -> `createdStr`

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -127,10 +127,10 @@ commandParser defaultNotesDir today = do
...
-      day <-
+      created <-
         option dayReader $
-          long "day"
-            <> metavar "DAY"
+          long "created"
+            <> metavar "CREATED"
             <> value today
             <> showDefault
             <> help "Zettel creation date in UTC"
```
srid, line 136: "Zettel creation date/time". And it should not say "UTC" anymore (since it isn't).

neuron/src/app/Neuron/Reader/Org.hs
```diff
@@ -36,16 +36,22 @@ parseOrg _ s = do
...
 extractMetadata doc
   | Just ((_, _, Map.fromList -> properties), _) <- getH1 doc = do
-    date <- traverse parseDate $ lookup "date" properties
+    createdWithTime <- traverse parseDate $ lookup "created" properties
+    createdDate <- traverse parseDate $ lookup "date" properties
+    let created =
+          case createdWithTime of
+            Nothing -> createdDate
+            _ -> createdWithTime
+
```
srid, line 45: Refactor this using `<|>`. cf. https://www.parsonsmatt.org/2016/11/18/clean_alternatives_with_maybet.html

neuron/src/app/Neuron/Reader/Org.hs
```diff
@@ -36,16 +36,22 @@ parseOrg _ s = do
...
     let title = Nothing
         tags = fmap Tag . words <$> lookup "tags" properties
         unlisted = parseUnlisted <$> lookup "unlisted" properties
     pure $ Just Meta {..}
   | otherwise = pure Nothing
   where
-    parseDate :: Text -> Either ZettelParseError Day
-    parseDate date = maybeToRight (Tagged $ "Invalid date format: " <> date) $ parseZettelDate @Maybe date
+    parseDate :: Text -> Either ZettelParseError LocalTime
+    parseDate date = maybeToRight (Tagged $ "Invalid date/time format: " <> date) $ parseZettelDate @Maybe date
```
srid, line 54: rename `date` -> `datetime`

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -35,7 +36,11 @@ instance FromYAML Meta where
         <$> m .:? "title"
         -- "keywords" is an alias for "tags"
         <*> (liftA2 (<|>) (m .:? "tags") (m .:? "keywords"))
-        <*> m .:? "date"
+        <*> (m .:? "created")
+          .!= ( case parseEither (m .:? "date") of
+                  Left _ -> Nothing
+                  Right v -> v
+              )
```
srid, line 43: See comment about `<|>` above.

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -45,25 +50,38 @@ instance FromYAML Meta where
...
 
--- | The format in which we decode and encode zettel dates.
-zettelDateFormat :: String
-zettelDateFormat = "%Y-%m-%d"
+-- | The default format in which we decode and encode zettel dates.
+defaultZettelDateFormat :: String
+defaultZettelDateFormat = "%Y-%m-%dT%H:%M"
+
+-- | All formats of user-edited tags 'created', that can be parsed.
+zettelDateFormats :: [String]
```
srid, line 69: use `NonEmpty String` (already imported in Relude)

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -45,25 +50,38 @@ instance FromYAML Meta where
...
+zettelDateFormats = ["%Y-%m-%d", defaultZettelDateFormat]
+
+formatZettelLocalTime :: LocalTime -> Text
+formatZettelLocalTime =
+  toText . formatTime defaultTimeLocale defaultZettelDateFormat
 
 formatZettelDate :: Day -> Text
 formatZettelDate =
-  toText . formatTime defaultTimeLocale zettelDateFormat
+  toText . formatTime defaultTimeLocale "%Y-%m-%d"
```
srid, line 78: put `"%Y-%m-%d"` in a variable, `defaultZettelDayFormat`.

and rename `defaultZettelDateFormat` to `defaultZettelDateTimeFormat`

srid: Actually, just create an ADT:

```haskell
data ZettelDateFormat
  = ZettelDateFormat_Day
  | ZettelDateFormat_DateTime

zettelDateFormat :: ZettelDateFormat -> String 
zettelDateFormat = \case 
  ZettelDateFormat_Day -> "%Y-%m-%d"
  ...
```

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -45,25 +50,38 @@ instance FromYAML Meta where
...
+
+parseZettelDate :: (MonadFail m, Alternative m) => Text -> m LocalTime
+parseZettelDate t =
+  let s = toString t
+      pars = parseDateM s
+   in asum $ pars <$> zettelDateFormats
```
srid, line 84: get rid of the `let` binding

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -45,25 +50,38 @@ instance FromYAML Meta where
...
 
-parseZettelDate :: MonadFail m => Text -> m Day
-parseZettelDate =
-  parseTimeM False defaultTimeLocale zettelDateFormat . toString
+parseDateM :: MonadFail m => String -> String -> m LocalTime
+parseDateM s = \fmt -> parseTimeM False defaultTimeLocale fmt s
```
srid, line 87: `parseDateM fmt = ...` and then `asum $ (flip parseDateM $ toString t) <$> zettelDateFormats` above.

neuron/src/lib/Neuron/Zettelkasten/Zettel/Parser.hs
```diff
@@ -37,14 +38,14 @@ parseZettel format zreader fn zid s = do
               ((,True) . plainify . snd <$> getH1 doc)
                 <|> ((,False) . takeInitial . plainify <$> getFirstParagraphText doc)
           tags = fromMaybe [] $ Meta.tags =<< meta
-          day = case zid of
+          created = case zid of
             -- We ignore the "data" meta field on legacy Date IDs, which encode the
             -- creation date in the ID.
-            ZettelDateID v _ -> Just v
-            ZettelCustomID _ -> Meta.date =<< meta
+            ZettelDateID v _ -> Just $ LocalTime v midnight
```
srid, line 44: I'd use `midday` here.

srid:
About the name `created` - before we finalize on it, what do the other Zettelkasten apps use?

clojj:
timeline looks like this... to be improved I think
neuron-timeline

neuron/src/app/Neuron/Reader/Org.hs
clojj: oh my... I was searching for exacty that, cool

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
clojj: yep.. I knew there was a 'flip' in it :)

clojj:
> About the name `created` - before we finalize on it, what do the other Zettelkasten apps use?

yes, good idea to look for other alternatives first

clojj:
also, if I find some time, I'd like to add some tests too

srid:
I thought about this change for a bit, and I think we should make these final changes:

* [ ] Drop `created` terminology and retain `date`. Rationale here: https://github.com/srid/neuron/issues/257#issuecomment-672891322
* [ ] Change the Zettel field to `zettelDate :: Either Day LocalTime`. We should not automatically add a time (midnight or midday) if the user has not specified one. Less magic, and more explicit types are a good thing.
* [ ] Revert the visual regression to how `?timeline` gets rendered; i.e., it should display only day, not the time. However, make it display the full time of the day (if specified) on mouse hover (by setting the `title` HTML attribute). 
* [ ] Add tests

clojj:
>     * [ ]  Change the Zettel field to `zettelDate :: Either Day LocalTime` (which incidentally is isomorphic to `(Day, Maybe TimeOfDay)`
I think that is an excellent idea !
As for the isomorphism... so the Either type is easier to handle, right ?

srid:
@clojj I think Either is fine.

clojj:
Do we need to define this instance then ?
```
instance ParseTime DateMayTime where
  substituteTimeSpecifier _ = 
  parseTimeSpecifier _ = 
  buildTime l xs = 
```
the time package doesn't seem to export this typeclass though (?)

clojj:
ok, sorry... I should first look at that type signature, will try to glue this Either together...

clojj:
@srid closing this PR in favor of another branch... will create the new PR soonish

Issue #326, opened 2020-08-06
Hide footnote refs from search result preview


srid:
Use `elNoSnippetSpan` from https://github.com/srid/neuron/commit/424c0c80595c02b6ec81873f8d8ea49de6d5f76b on footnote refs. Should be done in https://github.com/srid/reflex-dom-pandoc

---

![image](https://user-images.githubusercontent.com/3998/89476292-54eaf680-d758-11ea-99ad-5042f60a35fa.png)

PR #337, opened 2020-08-13
Change zettel ::after into a superscript

jared-w:
after:

### DIFF
--- a/neuron/src/lib/Neuron/Web/Query/View.hs
+++ b/neuron/src/lib/Neuron/Web/Query/View.hs
@@ -121,7 +121,7 @@ renderZettelLink conn (fromMaybe def -> linkView) Zettel {..} = do
             else Just $ "Tags: " <> T.intercalate "; " (unTag <$> zettelTags)
     elAttr "span" ("class" =: "zettel-link" <> withTooltip linkTooltip) $ do
       neuronRouteLink (Some $ Route_Zettel zettelID) mempty $ text zettelTitle
-      el "sup" $ text "ᛦ"
+      el "sup" $ do elAttr "span" noSnippet $ text "ᛦ"
   where
     -- Prevent this element from appearing in Google search results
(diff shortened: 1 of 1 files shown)


neuron/src/lib/Neuron/Web/Query/View.hs
```diff
@@ -121,6 +121,7 @@ renderZettelLink conn (fromMaybe def -> linkView) Zettel {..} = do
             else Just $ "Tags: " <> T.intercalate "; " (unTag <$> zettelTags)
     elAttr "span" ("class" =: "zettel-link" <> withTooltip linkTooltip) $ do
       neuronRouteLink (Some $ Route_Zettel zettelID) mempty $ text zettelTitle
+      el "sup" $ text "ᛦ"
```
srid, line 124: Can this be done in CSS `content`? I imagine you have an empty `sup` element, and style the content in CSS ... 

Otherwise you would have to use `noSnippet` from #326 in <sup> to suppress it from search preview.

jared-w: I experimented with doing it in CSS. I found it unsatisfactory trying to get the sizing and positioning correct in a way that behaved well if the browser screen was resized. Doing it in the CSS also prevents it from being highlighted if someone copies the text (can be a pro or a con depending on your view. I personally liked it being physically in the text)

Really, though, the main consideration for having it in the HTML was that it was easier to implement and is more easy to make accessible to screen readers. Had to get the noSnippet by placing the text inside an inner span because google only supports noSnippet on `div`, `section`, and `span`.

srid:
Regression.

image

srid:
@jared-w I had to revert this, because there is a bug. The suffix should appear only on folgezettel links. Note the `folgezettel` CSS class that normally takes care of it. Feel free to open a separate PR if you get to resolving it.

jared-w:
Huh, that's embarrassing. I could've sworn I tested exactly that earlier... I appreciate the swift response. I'll look at it this weekend if I get the chance.

Issue #222, opened 2020-06-09
Using a subset as blog


sureyeaah:
So I've been using neuron for a few days now and it seems perfect for note taking. I want to use a subset of my notes as a static blog. Can neuron do that currently? I would be happy to make any PRs necessary to support this.

srid:
This is exactly what I do in [srid.ca](https://www.srid.ca/) ([source](https://github.com/srid/srid.ca)). Basically [tag](https://neuron.zettel.page/2011505.html) the subset of notes with something like `timeline` and then pull it in the blog index page using the [link query](https://neuron.zettel.page/2011506.html) `<z:zettels?tag=timeline&timeline&cf>`. 

One thing that's missing in blog support is #73

sureyeaah:
Yeah that's what I was trying to do as well but how do you publish only those notes that have that tag?

srid:
Okay, you are asking about generating static site from a *subset* of your zettelkasten set. What defines a subset, though? A zettelkasten is densely linked; if you say that [A, B, C] is the subset, what happens if note B links to note D? B will include the title of the note D, thereby leaking private note data into the subset.

srid:
Alslo consider the case of B doing a link query using, say, `<z:zettels?tag=foo>`. If D, E, F are tagged "foo", then B would include a link to the zettels outside the subset. Those links would be broken, in addition their titles leaking into the published site.

I thought about this problem briefly before (but did not brainstorm in detail), and ended up keeping my private zettelkasten separate from the public one.

sureyeaah:
What if we let each note be marked as public/private and during generation, if we specify  that we only want the public ones, we only generate the public ones and in case a private note is linked from a public one - throw a warning/error or replace the link with a filler text or link to a 404 page?

srid:
We can use tags (say tag `blog`) to mark notes, and then perhaps:

```
neuron rib -o ~/www/blog.myself.com --only-tag=blog
```

Basically, *only* include zettels tagged `blog` during graph building and skip *everything else*. That might work.<sup>[1]</sup>

We don't need 404 page, because errors will be inlined [like this](https://github.com/srid/neuron/issues/201#issuecomment-633062878).

---
<sup>[1] though I personally wouldn't use it, so as to not leak private stuff by mistake!</sup>.

srid:
Might also have to do #141 in the scenario of untagging a published zettel.

leira:
[...]

Issue #50, opened 2020-04-03
Are cross links possible ?


573:
Will it be possible to reference other zettelkastens using the z: protocol ? I. e. using
```
[2011406](z://neuron.srid.ca)
```
instead of [Creating and Editing zettels 2011406](https://neuron.srid.ca/2011406.html).

srid:
In order to visually distinguish regular z: links from cross links, the later would probably look like this:

```markdown
[neuron.srid.ca/2011406](z:)
```

And rendered in HTML using "neuron.srid.ca/2011406" (in place of the bare ID) as the link text.

---

But I fail to see the value of adding this feature. What's the real-world use case for linking between Zettelkastens?

573:
Nevermind, just thought about having a convenient way to refer to other peoples' zettelkastens.
Maybe that is out-of-scope of the zettelkasten idea anyways.

srid:
I'm opening this, because I find myself wanting this feature lately. :-)

If I add a link an external zettelkasten like this:

```markdown
Users are delighted of it: <https://haskell.zettel.page/2016101.html>
```

It currently renders like this:

![image](https://user-images.githubusercontent.com/3998/79877459-40a60800-83ba-11ea-97ce-d161d6d07594.png)

But it would be great if it rendered like:

![image](https://user-images.githubusercontent.com/3998/79877618-71863d00-83ba-11ea-87b0-5bd522abe035.png)

dpwiz:
I would suggest [InterWiki](https://en.wikipedia.org/wiki/Interwiki_links)-style links, where each server is responsible for its map.

This would allow for [shorter](https://en.wikipedia.org/wiki/Petname) and more stable cross-linking.
I.e. somebody migrates their ZK to a new domain and everybody else is left with either broken links or search-and-replace task on their hands.

In addition, this would allow some server-to-server functionality later. Maybe even [Fediverse](https://en.wikipedia.org/wiki/Fediverse) integration.

dpwiz:
And the maps don't have to be `http`! One can even uplift their instance to IPFS/GNUnet namespaces, while keeping notes intact and content-addressable for everybody.

srid:
@dpwiz Interesting ideas. Today I can imagine doing the following:

1. Add an "otherNeuronSites" key to `neuron.dhall` where you specify a mapping of `$siteName` to `$siteBaseUrl`. Like InterWiki configuration.

[...]

PR #339, opened 2020-08-14
Multi site RFC

sureyeaah:
[**Rendered**](https://github.com/sureyeaah/neuron/blob/multi-site/guide/rfc-multi-site.md#generating-multiple-sites)

Addresses #222

### DIFF
--- a/guide/rfc-multi-site.md
+++ b/guide/rfc-multi-site.md
@@ -31,7 +31,7 @@ To configure multiple sites, we modify the existing configuration format.
 For backwards compatability, in absence of the `sites` field, we conside a single site implictly named `default`. All fields that are allowed at the top level (in the new config format) as well as for individual site config can be used here.
 
 ## Example
-```json
+```dhall
 { minVersion = "0.4.0.0"
 , sites =
   { default =
--- a/guide/rfc-multi-site.md
+++ b/guide/rfc-multi-site.md
@@ -1,0 +1,52 @@
+# Generating multiple sites
+
+In order to support subset as a blog (see [#222](https://github.com/srid/neuron/issues/222)), we need the ability to configure multiple sites (e.g. one for your local notes and another for your blog) as well as the ability to publish only a subset of the Zettelkasten (e.g. your blog might only publish zettels tagged `blog`).
+
+## Filtering zettels while generating
+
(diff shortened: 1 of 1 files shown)


srid:
BTW, table rendering is broken: https://github.com/sureyeaah/neuron/blob/multi-site/guide/rfc-multi-site.md#generating-multiple-sites

GitHub [does support it](https://docs.github.com/en/github/writing-on-github/organizing-information-with-tables), so I wonder what could be wrong. Perhaps something to do with newlines?

srid:
A third way of solving this is to consider #50 but in local context. Suppose there are two *local* zettelkastens -- private & public -- we could make it possible for one zettelkasten to (cross) link to another (but *not* vice versa). So "public" zettelkasten would be independent; but the "private" zettelkasten can cross link to the "public" zettels.  Coming up with a spec to make this work doesn't seem exactly easy though.

srid:
Thinking more about _cross-linking_ ... instead of having sub-zettelkastens, we now think in terms of _composing_ (otherwise independent) zettelkastens.

So, if you have a zettelkasten called "public" -- and it exists as a sub-directory (or git submodule) as ./public/neuron.dhall -- then you can have another zettelkasten's neuron.dhall "point" to it as `{ "public" = ./public }`. This another zettelkasten can now link to the public zettelkasten using an ID prefix. For example, to link to ./public/foo.md - you would write `<public:foo>`. Its graph would include the zettels from "public" but with this ID prefix. The generated HTML would go in a sub-directory, i.e., `public/foo.html`.

This also solves the problem of public/index.html being in conflict with the master zettelkasten's zettel of same ID.

And it solves the constraint of "public" zettelkasten being allowed only to link to itself (unless it is configured to have cross-links)

sureyeaah:
How does this affect querying? If I query by a tag, is it supposed to query cross-links as well or is there a special syntax for choosing an ID prefix?

srid:
Since both the zettelkastens will be part of the same graph, if you say query by tag - it will include zettels from both of them.

To be more precise, it is not exactly cross-linking - rather, more of "embedding" another zettelkasten into your graph ... but with IDs namespaced. Here, you would be embedding your blog zettelkasten onto your private zettelkasten.

srid:
Closing in favour of #355

PR #343, opened 2020-08-15
Allow optional time in `date` metadata field

### LINKED ISSUE #257: Extend Metadata 'date' to include hours/minute/seconds ?

clojj:
Fixes #257 

TODOs:
- use Alternative for that ugly code in `parseZettelDate` ?
- unit tests


clojj:
> I thought about this change for a bit, and I think we should make these final changes:
> 
>     * [ ]  Drop `created` terminology and retain `date`. Rationale here: [#257 (comment)](https://github.com/srid/neuron/issues/257#issuecomment-672891322)
> 
>     * [ ]  Change the Zettel field to `zettelDate :: Either Day LocalTime` (which incidentally is isomorphic to `(Day, Maybe TimeOfDay)`. We should not automatically add a time (midnight or midday) if the user has not specified one. Less magic, and more explicit types are a good thing.
>       
>       * Perhaps make a type alias first: `type DateMayTime = Either Day LocalTime`
> 
>     * [ ]  Revert the visual regression to how `?timeline` gets rendered; i.e., it should display only day, not the time. However, make it display the full time of the day (if specified) on mouse hover (by setting the `title` HTML attribute).
> 
>     * [ ]  Add tests
>       
>       * Make sure that zettel sorting takes time into account, if specified.
> 
> 
> EDIT:
> 
>     * [ ]  Remove the "Created on:" at the bottom, because that can be confusing if date is not always creation date.

...posted here, just to pick up from the last PR

clojj:
I guess you were referring to that sorting test in ZettelSpec ?

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -48,22 +56,39 @@ instance FromYAML Meta where
...
+    Left day -> formatTime defaultTimeLocale dateFormat day
+    Right localtime -> formatTime defaultTimeLocale dateTimeFormat localtime
+
+formatZettelDateAsDay :: DateMayTime -> Text
+formatZettelDateAsDay =
+  toText . \case
+    Left day -> fmt day
+    Right localtime -> fmt (localDay localtime)
+  where
+    fmt = formatTime defaultTimeLocale dateFormat
```
srid, line 79: Replace where clause with point-free style:

```haskell
formatTime defaultTimeLocale dateFormat . toText . \case ...
```

neuron/src/lib/Neuron/Zettelkasten/Zettel/Parser.hs
```diff
@@ -37,14 +38,14 @@ parseZettel format zreader fn zid s = do
               ((,True) . plainify . snd <$> getH1 doc)
                 <|> ((,False) . takeInitial . plainify <$> getFirstParagraphText doc)
           tags = fromMaybe [] $ Meta.tags =<< meta
-          day = case zid of
+          date = case zid of
             -- We ignore the "data" meta field on legacy Date IDs, which encode the
             -- creation date in the ID.
-            ZettelDateID v _ -> Just v
+            ZettelDateID v _ -> Just (Right $ LocalTime v midday)
```
srid, line 44: Why are we automatically guessing a time (`midday`) here instead of using the `Left` day?

neuron/test/Neuron/Reader/OrgSpec.hs
```diff
@@ -0,0 +1,38 @@
...
+
+itParsesDate :: String -> Text -> SpecWith ()
+itParsesDate name s =
+  it name $ do
+    parseDate s `shouldBe` Right (Right (LocalTime (ModifiedJulianDay 59077) (TimeOfDay 9 42 0)))
+
+itParsesInvalid :: String -> Text -> SpecWith ()
+itParsesInvalid name s =
+  it name $ do
+    parseDate s `shouldBe` Left (Tagged $ "Invalid date format: " <> s)
```
srid, line 38: Don't create functions when not necessary. It is okay to be verbose and violate DRY. See https://www.srid.ca/47e0e49a.html

Just put everything in `spec`.

neuron/test/Neuron/Zettelkasten/ZettelSpec.hs
```diff
@@ -24,21 +26,36 @@ spec = do
       noContent = MetadataOnly ()
   describe "sortZettelsReverseChronological" $ do
     let mkDay = fromGregorian 2020 3
+        mkZettelDay = Just . Left . mkDay
+        mkZettelLocalTime day hh mm = (Just (Right $ (LocalTime (mkDay day) (TimeOfDay hh mm 0))))
```
srid, line 30: Prefer `.` or `$` over nested parenthesis. i.e., `Just $ Right $ LocalTime (mkDay day) $ TimeOfDay ...`

neuron/test/Neuron/Zettelkasten/ZettelSpec.hs
```diff
@@ -24,21 +26,36 @@ spec = do
...
+      sortZettelsReverseChronological zs
+        `shouldBe` reverse zs
+
+    it "sorts correctly with localtime" $ do
+      let zs = [mkZettel 3 (mkZettelLocalTime 3 9 59) 2, mkZettel 3 (mkZettelLocalTime 3 10 0) 1]
+      sortZettelsReverseChronological zs
+        `shouldBe` reverse zs
+
+    it "sorts correctly with mixed dates" $ do
+      let zs = [mkZettel 3 (mkZettelDay 3) 2, mkZettel 3 (mkZettelLocalTime 3 0 0) 1]
```
srid, line 59: Could you use a different `day` value in the test case here?

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -46,7 +46,7 @@ data App = App
 data NewCommand = NewCommand
   { title :: Maybe Text,
     format :: Maybe ZettelFormat,
-    day :: Day,
+    localtime :: LocalTime,
```
srid, line 49: Call this `date :: LocalTime`. Since that's what it is called in the markdown metadata.

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -48,22 +56,39 @@ instance FromYAML Meta where
...
+
+parseZettelDate :: (MonadFail m, Alternative m) => Text -> m DateMayTime
+parseZettelDate t =
+  let s = toString t
```
srid, line 83: You can avoid the let binding using ViewPatterns:

```haskell
parseZettelDate (toString -> s) = ...
```

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -48,22 +56,39 @@ instance FromYAML Meta where
 --         "date" .= date
 --       ]
 
-instance FromYAML Day where
+instance FromYAML (Either Day LocalTime) where
```
srid, line 59: `instance FromYAML DayMayTime`?

neuron/src/lib/Neuron/Zettelkasten/Zettel/Meta.hs
```diff
@@ -48,22 +56,39 @@ instance FromYAML Meta where
...
   parseYAML =
     parseZettelDate <=< parseYAML @Text
 
-instance ToYAML Day where
+instance ToYAML (Either Day LocalTime) where
```
srid, line 63: `instance ToYAML DayMayTime`?

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -145,7 +145,7 @@ commandParser defaultNotesDir today = do
           <|> fmap
             (const . Some . IDSchemeCustom)
             (option str (long "id" <> help "Use a custom ID" <> metavar "IDNAME"))
-      pure $ New $ NewCommand title format day (idSchemeF day) edit
+      pure $ New $ NewCommand title format now (idSchemeF day) edit
```
srid, line 148: This looks like it is discarding the date passed to `neuron --new --day=...`.

clojj:
@srid all comments resolved

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -145,7 +145,7 @@ commandParser defaultNotesDir now = do
           <|> fmap
             (const . Some . IDSchemeCustom)
             (option str (long "id" <> help "Use a custom ID" <> metavar "IDNAME"))
-      pure $ New $ NewCommand title format now (idSchemeF day) edit
+      pure $ New $ NewCommand title format (LocalTime dayParam midday) (idSchemeF dayParam) edit
```
srid, line 148: We shouldn't be automatically filling in the time at all, if the user has not specified one. 

I think there shoud be a `neuron new --date=foo` (replacing `--day`) where "foo" can be either date or a datetime. You can reuse the `DateMayTime` type alias here, and use `Either` to capture the sum.

clojj: doing this, options-lib displays this default though:
option

clojj: `      dateParam <-
        option dateReader $
          long "date"
            <> metavar "DATE"
            <> value (Left today)
            <> showDefault
            <> help "Zettel creation date/time"
`

clojj: I guess I have to change "showDefault" ?

clojj: Ok, have foung it: `showDefaultWith`

clojj:
The JSON API has to be changed or is that wip ?

neuron/test/Neuron/Reader/OrgSpec.hs
```diff
@@ -0,0 +1,33 @@
...
+        it name $ do
+          parseDate s `shouldBe` Right (Right (LocalTime (ModifiedJulianDay 59077) (TimeOfDay 9 42 0)))
+      itParsesInvalid name s =
+        it name $ do
+          parseDate s `shouldBe` Left (Tagged $ "Invalid date format: " <> s)
+
+  describe "date-tag parsing" $ do
+    itParsesDay "with day" "2020-08-16"
+    itParsesDate "with localtime" "2020-08-16T09:42"
+    itParsesInvalid "with invalid" "2020-08-16 09:42"
```
srid, line 33: Don't create functions or let bindings here. Just inline them. 

> As a rule, allow yourself to write the duplicate code at least 3 times before considering abstracting it. <https://www.srid.ca/47e0e49a.html>

Here, these functions are used exactly once (not 3 times).

neuron/src/app/Neuron/CLI/Types.hs
```diff
@@ -127,13 +127,13 @@ commandParser defaultNotesDir today = do
...
-          long "day"
-            <> metavar "DAY"
-            <> value today
-            <> showDefault
-            <> help "Zettel creation date in UTC"
+      dateParam <-
+        option dateReader $
+          long "date"
+            <> metavar "DATE"
+            <> value (Left today)
```
srid, line 134: The default should be `Right now`, I think? I imagine we are making `neuron new` inject the full datetime by default in the generated zettel note.

srid:
> The JSON API has to be changed or is that wip ?

That is automatically taken care of by the Generic Aeson instance, when you modified the `Zettel` type by changing the field to `zettelDate :: Maybe DateMayTime`. I added the label only as a reminder for myself, in the context of Cerveau (which stores the JSON in postgres, and will require migration/reset when upgrading production)

Issue #312, opened 2020-07-29
Flexible links


srid:
**Proposal**

I propose the following linking syntax (retaining the underlying link URI):

```markdown
This is current style, and will continue to be supported: <foo>

Proposed default: [[foo]]

Flexible link, that allows setting inner text: [[foo | the foo note]].
```

This does a couple of things:

1. Makes the already supported `[[..]]` style links the *default*. 
2. Supports overriding the link text, as discussed in #306 

Advance linking syntax would look the same, except for the change in the 'wrapper' syntax:

```markdown

Current: 
- <z:zettels?tag=haskell>
- <foo?cf>

Proposed: 
- [[z:zettels?tag=haskell]]
- [[foo?cf]] or [[foo?cf | a foo note]]
```

Let's use this opportunity to also support regular Markdown links:

```markdown
Plain markdown links: 
- [a foo note](foo.md)
- [a foo note](z:/foo)
```

Something to think about: 
- how would these links look and behave in other Zettelkasten apps (emacs, vimwiki, Obsidian #229, Zettlr, ..)? Do we care? @vcavallo @felko @SquareBottle @EyebrowHairs
- How should links with user-defined inner text render stylistically?

---

### Final Proposal and Progress

- [x] Use `[[..]]` and `[[[...]]]` instead of `<..>` #351
- [ ] Allow overriding link text in wikilinks, eg: `[[ foo | a foo note ]]`
- [x] Regular Markdown links: `[a foo note](foo.md)`

srid:
Oops, forgot to add one more thing:

```markdown
Plain markdown links: 
- [a foo note](foo.md)
- [a foo note](z:/foo)
```

vcavallo:
As far as I can tell so far, unfortunately a lot of these still won't work with _vimwiki_ because it will assume that the bit within the `()` is a _filename_ and it will fail upon adding any query parameters or anything else.

This proposal will still be an improvement, though, as far as rendering the markup on the web.

EyebrowHairs:
[...]

PR #351, opened 2020-08-20
Wiki-style links

srid:
Part of #312

Replace autolinks with Wiki links (see below), making them the default. Autolinks will be supported for backwards compat.

- `<foo?cf>` replaced by `[[foo]]`
- `<foo>` replaced by `[[[foo]]]` 
- `<z:zettels?tag=xyz>` replaced by `[[[z:zettels?tag=xyz]]]`

In particular note that `[[[...]]]` makes the link a folgezettel, whereas `[[..]]` makes it a cf. This applies to full URI queries too.

The following are outside the scope of this PR:

- `{{..}}` style links
- Plain markdown links

Tasks,

- [x] Working prototype
- [ ] Complete, and refactor, implementation
- [ ] Add unit test coversage
- [ ] Replace all links in ./guide, and update documentation

### DIFF
--- a/guide/2011502.md
+++ b/guide/2011502.md
@@ -5,7 +5,7 @@ tags:
 
 # Tutorial
 
-Make sure you have already installed neuron (see <2011501?cf>). Then follow this tutorial to get your own Zettelkasten up and running.
+Make sure you have already installed neuron (see [[2011501]]). Then follow this tutorial to get your own Zettelkasten up and running.
 
 ## Test with an existing Zettelkasten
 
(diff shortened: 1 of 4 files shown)


vcavallo:
In case it's helpful for anyone in the future, these `sed` snippets should help with converting your existing `<asdf.md?cf>` and `<asdf.md>` style links to the new bracket-style. These worked for me but are largely untested, so be careful, backup your data, and your mileage may vary:

```
# use from within the directory where your .md files reside.
# for ?cf links:
sed -E "s:<([a-zA-Z0-9.]*)\?cf>:[[[\1]]]:g" *.md

# for standard links:
# sed -E "s:<([a-zA-Z0-9.]*)>:[[\1]]:g" *.md
```

These will also change any other text that is wrapped in `<   >`, so beware of that, as well.

PR #355, opened 2020-08-20
RFC: cross links

sureyeaah:
[**Rendered**](https://github.com/sureyeaah/neuron/blob/cross-links/guide/rfc-cross-links.md#cross-linking)

Cross-linking feature as suggested by @srid in #339 

Addresses #222

### DIFF
--- a/guide/rfc-cross-links.md
+++ b/guide/rfc-cross-links.md
@@ -1,0 +1,24 @@
+# Cross-linking
+
+Cross linking lets you compose zettelkastens via "namespaced links". Consider these scenarios:
+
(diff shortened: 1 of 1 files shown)


srid:
The spec should support regular Markdown links, eg: `[a link](../work/project1.md)`

srid:
I'm going to close this. Neuron v2 is focused on being as simple as possible, shifting any sophisticated features to the plugin layer. We can revisit cross-links in the future. The blog subset requirement, which this spec was originally meant to address, can be implemented in simpler ways (https://github.com/srid/neuron/discussions/222#discussioncomment-248875).

PR #370, opened 2020-09-04
Advance nixpkgs, for HLS 0.3

srid:
https://github.com/haskell/haskell-language-server/blob/master/ChangeLog.md#030

Looking at the ChangeLog, 0.4 is already released. Wonder how long before nixpkgs gets it ...


srid:
https://github.com/NixOS/nixpkgs/pull/97062

Issue #383, opened 2020-09-09
[Feature Request] Inserting custom HTML into static web site generation


leonardt:
Hi there, thanks for your great work on this project!

I'm using this as a part of my personal website, and I have a top-bar linking to it as a "notes" section.  I'd love to be able to include the same top bar in the generated static web site, that way I have a consistent "theme" (mainly I want to have a button back to the homepage that links to a portion of the website outside of the statically generated content from neuron.

I'd be happy to look into implementing this myself, but was hoping I could get a pointer to where in the code I could insert the logic.  Also I'm open to suggestions on how to integrate this from a user perspective in an extensible way (ideally beyond my use case).  For example, do we simply provide some configuration option for some static HTML snippets to insert? This would cover my use case.  A more general solution might be to provide some hooks for the user to provide arbitrary code to insert content during the page generation process.  This seems more useful in a broader context, but not sure how hard this is to achieve in the existing code.

srid:
@leonardt See #253 (for general theming).

What you are asking is simliar to #362.

We want to figure out a simpler solution, without adding too many configuration points.

See here for all theming related requests: https://github.com/srid/neuron/labels/case-for-themes

srid:
> I'd be happy to look into implementing this myself, but was hoping I could get a pointer to where in the code I could insert the logic.

https://github.com/srid/neuron/blob/d3e69a91b5542ec47eeee733b6e6d4e015c55546/neuron/src/app/Neuron/Web/View.hs#L99-L128

Doing it in a general way is gonna be a challenge, as we use Haskell for UI.

vizziv:
I started looking into #362 myself, but I wasn't able to figure out how to insert arbitrary HTML using Reflex.

Would a possible first-draft solution to this and #362 be to allow adding up to one `script` tag to the `head`? (Possibly but not necessarily replacing the MathJax script.) A single `script` could use JavaScript to add more elements (including loading more JS and CSS files) while leaving the Reflex code relatively clean.

More generally than adding one `script`, one could imagine allowing a list of `script`s and a list of CSS `link`s. But there are so many attributes for each such tag (`async`, `deferred`, `integrity`, etc.) that the dhall configuration would get unwieldy.

The single-`script` solution wouldn't necessarily be as fast as the multiple-`script`/`link` solution, but it's simpler to configure and would have perfectly acceptable performance for a personal zettelkasten.

srid:
[...]

Issue #362, opened 2020-08-27
Allow custom <head> 


srid:
Allow users to specify an optional fragment of HTML that should go in the `<head>` element of all generated pages (in addition to neuron's builtin).

```
# If this file exists, inject it to <head> of all pages
./neuron/head.html
```

The `./neuron/head.html` file, if the user has specified one, can inject any custom CSS or JS includes.

If the user has not specified one, neuron will use its default custom head which is set to the current mathjax include (tex-mml-chtml.js). Note that semantic UI, google fonts, etc. are always included (they are part of builtin includes).

srid:
And let's use this opportunity to deprecate `mathJaxSupport` from neuron.dhall.

srid:
Done in #385.

Issue #361, opened 2020-08-26
Enable different MathJax URL


vizziv:
Right now the URL to the MathJax Javascript file is hard-coded, but there are several possible options one might want to use (e.g. to use SVG rendering instead of CHTML).

One way to support this would be to switch the `mathJaxSupport` Dhall configuration option from a bool to an optional URL.

vizziv:
Note: this is related to #212, but I think this is much simpler to implement.

More generally, one might imagine giving the user more control over what lives in the `<head/>` of their pages, e.g. to load custom scripts. One can use such a script to solve the repeated `\newcommand` problem brought up in #348.

srid:
> More generally, one might imagine giving the user more control over what lives in the <head/> of their pages, e.g. to load custom scripts. 

I like this. Opened #362.

vizziv:
This seems subsumed by #362, so closing.

Issue #212, opened 2020-05-29
Switch to KaTeX


felko:
Neuron currently uses MathJax. [KaTeX](https://github.com/KaTeX/KaTeX) appears to be faster, and supports pre-rendering of TeX expressions in plain HTML, which might be interesting for offline browsing, as long as the output size is reasonable (unless #96 is implemented, which would bundle KaTeX with neuron anyway).
KaTeX is already [supported in pandoc](https://hackage.haskell.org/package/pandoc-2.9.2.1/docs/Text-Pandoc-Options.html#t:HTMLMathMethod) so the implementation should be easy.

RyanGreenup:
It’s worth mentioning that KaTeX doesn’t have a way to automatically number Equations. 

But numbering equations with mathjax using `align `  is inconsistent with latex which can make pandoc tricky to use to get a PDF so katex might be better anyway.

srid:
Another reason to switch to KaTeX, for smoother math rendering in Cerveau: https://twitter.com/sridca/status/1304468929428021251

srid:
Users can use KaTex in #385. Might just have to document, if we aren't switching from MathJS to KaTeX by default.

srid:
Closing per https://neuron.zettel.page/next 

When using Emanote, you get to choose the Math library as it doesn't hardcode anything for you.

PR #385, opened 2020-09-11
Allow custom HTML in `<head>`.

vizziv:
Adds `headHtml : Optional Text` option to the configuration. It can be set in one of two ways, but not both simultaneously.

- The option can be set as usual in `./neuron.dhall`.
- If the file `./head.html` exists, then `headHtml` is set to `Some [head.html's contents]`.

Addresses #362 directly. Also solves or enables workarounds for several other issues (e.g. #212, #361, #383).

### DIFF
--- a/neuron/src/app/Main.hs
+++ b/neuron/src/app/Main.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -15,6 +16,7 @@ import Neuron.Config.Type (Config)
 import Neuron.Version (neuronVersion)
 import Neuron.Web.Generate (generateSite)
 import Neuron.Web.Generate.Route (staticRouteConfig)
+import Neuron.Web.HeadHtml (getHeadHtml)
 import qualified Neuron.Web.Manifest as Manifest
 import Neuron.Web.Route (Route (..), runNeuronWeb)
 import Neuron.Web.View (renderRoutePage)
@@ -32,13 +34,14 @@ generateMainSite config = do
   notesDir <- ribInputDir
   buildStaticFiles ["static/**", ".nojekyll"]
   manifest <- fmap Manifest.mkManifest $ getDirectoryFiles notesDir Manifest.manifestPatterns
+  headHtml <- getHeadHtml
   let writeHtmlRoute :: Route a -> (ZettelGraph, a) -> Action ()
       writeHtmlRoute r x = do
         html <- liftIO $
           fmap snd $
             renderStatic $ do
               runNeuronWeb staticRouteConfig $
-                renderRoutePage neuronVersion config manifest r x
+                renderRoutePage neuronVersion headHtml config manifest r x
         -- FIXME: Make rib take bytestrings
         writeRoute r $ decodeUtf8 @Text html
   void $ generateSite config writeHtmlRoute
--- a/neuron/src/app/Main.hs
+++ b/neuron/src/app/Main.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -41,7 +40,7 @@ generateMainSite config = do
           fmap snd $
             renderStatic $ do
               runNeuronWeb staticRouteConfig $
-                renderRoutePage neuronVersion headHtml config manifest r x
+                renderRoutePage neuronVersion config headHtml manifest r x
         -- FIXME: Make rib take bytestrings
         writeRoute r $ decodeUtf8 @Text html
   void $ generateSite config writeHtmlRoute
--- a/neuron/src/app/Neuron/Config.hs
+++ b/neuron/src/app/Neuron/Config.hs
@@ -13,11 +13,10 @@
 
 module Neuron.Config
   ( getConfig,
-    parseConfig,
+    parsePure,
   )
 where
 
-import Data.Set ((\\), member)
 import Data.Either.Validation (validationToEither)
 import Development.Shake (Action, readFile')
 import Dhall (FromDhall)
@@ -25,9 +24,8 @@ import qualified Dhall
 import qualified Dhall.Core
 import qualified Dhall.Parser
 import qualified Dhall.TypeCheck
-import qualified Dhall.Map
 import Neuron.Config.Orphans ()
-import Neuron.Config.Type (Config (..), configFile, configKeys, mergeWithDefault)
+import Neuron.Config.Type (Config, configFile, defaultConfig, mergeWithDefault)
 import Relude
 import Rib.Shake (ribInputDir)
 import System.Directory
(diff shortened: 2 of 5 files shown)


srid:
@vizziv Thanks! Taking a quick look at the diff, I see that this differs from #362 in two ways:

- Why the neuron.dhall configuration, over just using `head.html` if it exists (see [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration))?
- Why is the default mathjax script still included? I thought we would have a default snippet for `head.html` when it is not specified by the user, and that this snippet would contain the existing mathjax include. The idea being that if the user overrides head.html, they become responsible for including whatever math script (math or katex or whatever) they need.

vizziv:
The answer in both cases is that I chose a solution that made the fewest changes to the code, as I wanted to submit the minimally invasive patch that added the feature. But I'm happy to make both changes.

Regarding how to organize the code for the first change: should we use the `Config` type to store `head.html`'s contents, or should we create a new module and type to handle it? (Or should we just put the code reading `head.html` directly into `Main`?)

vizziv:
Another question about the second change: right now we replace `$`/`$` pairs with `\(`/`\)` pairs for MathJax's sake. Should disabling MathJax entail disabling this replacement?

srid:
Checkout how the [icon and manifest files](https://neuron.zettel.page/67c6f7a0.html) are handled for inspiration. They are loaded in src/app/Main.hs as `manifest` and passed around (along with `config`) as a function argument to `renderRoutePage`. You can do the same for `head.html`. And notice how `Neuron.Web.Manifest` handles the creation of the manifest type, including handing the default case - of using neuron logo as the favicon.

srid:
re: MathJax - I'm not sure what you mean; neuron doesn't do any special processing, aside from letting commonmark handle it (see `CE.mathSpec` in `Neuron.Reader.Markdown`). The `mathJaxSupport` config field should be ... made to be totally ignored in this PR (with a deprecation warning), and entirely removed one day, I think.

srid:
You can probably sneak in the head.html stuff in a new field of the `Neuron.Web.Manifest:Manifest` type, while renaming that module/type suitably (it won't just be a "manifest" anymore).

vizziv:
Ahh, I'd already made a new module for this, creatively named `HeadHtml`. But it could definitely be easily combined with (an appropriately renamed) `Manifest` either now or in the future.

vizziv:
> The `mathJaxSupport` config field should be ... made to be totally ignored in this PR (with a deprecation warning), and entirely removed one day, I think.

I've removed `mathJaxSupport` from `Config`. I wasn't sure how to best add a deprecation warning, so instead we throw an error. This also throws an error for other spurious fields in `neuron.dhall`.

I'm not sure how you want to update the guide to reflect these changes, so I'll leave that to you and others if that's okay.

> re: MathJax - I'm not sure what you mean; neuron doesn't do any special processing, aside from letting commonmark handle it (see `CE.mathSpec` in `Neuron.Reader.Markdown`).

Yeah, looks like the `$`/`$`-to-`\(`/`\)` conversion is determined by some combination of `commonmark-pandoc` and `pandoc`. Will table this for now.

neuron/src/app/Main.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE LambdaCase #-}
```
srid, line 2: This can be removed.

neuron/src/app/Main.hs
```diff
@@ -32,13 +34,14 @@ generateMainSite config = do
...
   manifest <- fmap Manifest.mkManifest $ getDirectoryFiles notesDir Manifest.manifestPatterns
+  headHtml <- getHeadHtml
   let writeHtmlRoute :: Route a -> (ZettelGraph, a) -> Action ()
       writeHtmlRoute r x = do
         html <- liftIO $
           fmap snd $
             renderStatic $ do
               runNeuronWeb staticRouteConfig $
-                renderRoutePage neuronVersion config manifest r x
+                renderRoutePage neuronVersion headHtml config manifest r x
```
srid, line 44: `config` should come before `headHtml` in arguments.

neuron/src/app/Neuron/Config.hs
```diff
@@ -50,8 +52,22 @@ parsePure :: forall a. FromDhall a => FilePath -> Text -> Either String a
...
+      let spuriousMessage =
+            "Configuration in neuron.dhall includes spurious fields: " <>
+            intercalate ", " (("`" <>) . (<> "`") . toString <$> toList spuriousFields) <>
+            "."
+          -- TODO: remove once
+          mathJaxMessage =
+            if "mathJaxSupport" `member` spuriousFields
+              then " To disable MathJax support, include a (possibly blank) `./head.html` file."
+              else ""
+      Left $ spuriousMessage <> mathJaxMessage
```
srid, line 73: 1. There should be a comment explaining what this piece of code does.

1. If spuriousFields is a list of length 1 containing just "mathJaxSupport" - it should be ignored, without returning an error. For other cases, it should error out. The idea is that we remove `mathJaxSupport` entirely, while still allowing its presence in older neuron.dhall (if only to ignore it). Running the new neuron on an older zettelkasten shouldn't fail.

neuron/src/app/Neuron/Config.hs
```diff
@@ -37,10 +39,10 @@ getConfig = do
...
   configVal :: Text <-
     liftIO (doesFileExist configPath) >>= \case
-      True -> do
-        mergeWithDefault . toText <$> readFile' configPath
+      True ->
+        toText <$> readFile' configPath
       False ->
-        pure defaultConfig
+        pure emptyConfig
   either fail pure $ parsePure configFile $ mergeWithDefault configVal
```
srid, line 46: Hmm, what's the reason for this change? Default config is not used anymore?

neuron/src/app/Neuron/Web/View.hs
```diff
@@ -49,16 +50,17 @@ import qualified Skylighting.Format.HTML as Skylighting
...
 -- | Render the given route
-renderRoutePage :: PandocBuilder t m => Text -> Config -> Manifest -> Route a -> (ZettelGraph, a) -> NeuronWebT t m ()
-renderRoutePage neuronVersion config manifest r val =
+renderRoutePage :: PandocBuilder t m => Text -> HeadHtml -> Config -> Manifest -> Route a -> (ZettelGraph, a) -> NeuronWebT t m ()
+renderRoutePage neuronVersion headHtml config manifest r val = do
+  el "!DOCTYPE html" $ pure ()
   elAttr "html" ("lang" =: "en") $ do
     el "head" $ do
-      renderRouteHead config manifest r val
+      renderRouteHead headHtml config manifest r val
```
srid, line 58: `config` should come before `headHtml` in arguments.

neuron/src/app/Neuron/Web/View.hs
```diff
@@ -85,7 +87,8 @@ renderRouteHead config manifest route val = do
       elAttr "link" ("rel" =: "stylesheet" <> "href" =: "https://cdn.jsdelivr.net/npm/fomantic-ui@2.8.5/dist/semantic.min.css") blank
       elAttr "style" ("type" =: "text/css") $ text neuronCss
       elLinkGoogleFonts neuronFonts
-      when (mathJaxSupport config) $
+      -- Include the MathJax script unless a custom head.html is provided.
+      renderHeadHtmlOr headHtml $
         elAttr "script" ("id" =: "MathJax-script" <> "src" =: "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js" <> "async" =: "") blank
```
srid, line 92: I suggest moving the default HTML to `HeadHtml.hs`, and thereby changing `renderHtmlOr headHtml (elAttr ...)` to `elPandocRaw (Format "html") headHtml`. Why? Because that's what we do with `Manifest` (the default is handled in the module itself).

neuron/src/lib/Neuron/Config/Type.hs
```diff
@@ -10,7 +10,8 @@
 module Neuron.Config.Type
   ( Config (..),
     configFile,
-    defaultConfig,
```
srid, line 13: ?

srid (changes requested):

srid:
About the deprecation warning .. okay, let's not worry about it; let's ignore mathjax silently if used by the user. I added some comments in your PR.

About the $ -> ( conversion, that happens here: https://github.com/srid/reflex-dom-pandoc/blob/259c7a881dd4ba69f82ac2fdec83ddf8f5ee0fc5/src/Reflex/Dom/Pandoc/Document.hs#L204

neuron/src/app/Neuron/Config.hs
vizziv: Ahh, the introduction of `emptyConfig` was more important in a previous iteration of the code. We can remove it.

With that said, the `Config` interface was a little strange before. Exporting both `defaultConfig` and `mergeWithDefault` felt like a bit of a code smell. And on the line after this one, we would sometimes end up merging two copies of `defaultConfig`.

Proposed solution: change the `Config` interface to export neither `emptyConfig` nor `defaultConfig`, but change `mergeWithDefault` to
```haskell
mergeWithDefault :: Maybe Text -> Text
mergeWithDefault = \case
  Nothing -> defaultConfig
  Just config -> defaultConfig <> " // " <> config
```
Then this line of code can be `pure Nothing`.

neuron/src/app/Neuron/Config.hs
vizziv: Okay, I'll add an explanation and a special case for when `mathJaxSupport` is the only spurious field.

(Also, the "TODO: remove once" comment was meant to continue: ... "mathJaxSupport has been deprecated for long enough")

neuron/src/lib/Neuron/Config/Type.hs
vizziv: See previous comment about `Config.Type`.

neuron/src/app/Neuron/Config.hs
srid: Sounds good, re: proposed solution.

neuron/src/app/Neuron/Config.hs
```diff
@@ -35,23 +37,43 @@ import System.FilePath
...
+    then pure config
+    else do
+      let spuriousMessage =
+            "Configuration in neuron.dhall includes spurious fields: " <>
+            intercalate ", " (("`" <>) . (<> "`") . toString <$> toList spuriousFields) <> "."
+          mathJaxMessage =
+            if "mathJaxSupport" `member` spuriousFields
+              then " To disable MathJax support, include a (possibly blank) `./head.html` file."
+              else ""
+      fail $ spuriousMessage <> mathJaxMessage
```
srid, line 79: Config parsing has gotten a bit complex. I'm wondering if we should just break backwards compact, and have users remove this field manually.

neuron/src/app/Neuron/Web/View.hs
```diff
@@ -49,16 +50,17 @@ import qualified Skylighting.Format.HTML as Skylighting
 import qualified Skylighting.Styles as Skylighting
 
 -- | Render the given route
-renderRoutePage :: PandocBuilder t m => Text -> Config -> Manifest -> Route a -> (ZettelGraph, a) -> NeuronWebT t m ()
-renderRoutePage neuronVersion config manifest r val =
+renderRoutePage :: PandocBuilder t m => Text -> Config -> HeadHtml -> Manifest -> Route a -> (ZettelGraph, a) -> NeuronWebT t m ()
+renderRoutePage neuronVersion config headHtml manifest r val = do
+  el "!DOCTYPE html" $ pure ()
```
srid, line 55: ?

srid (changes requested):
@vizziv You know what, I'm now not sure if it is so important to support backwards compact, especially if it makes config parsing rather complex. 

Most users I imagine won't have `mathJaxSupport = false` in their config. Those that do, can always remove it when next updating neuron.

So let's get rid of the config fields check entirely. Please undo the Config changes in your PR, except for removing `mathJaxSupport` field entirely from the Config type.

neuron/src/app/Neuron/Config.hs
srid: No need to make this change in this PR (see comment below).

neuron/src/app/Neuron/Web/View.hs
srid: btw, you can use `blank` in place of `pure ()`.

vizziv: Oh, whoops, I had forgotten about this DOCTYPE addition. Handling DOCTYPE could be its own PR, in which case I'll remove this, but I also suspect that `<!DOCTYPE html>` is the right thing for most people.

The reasoning for adding it was for loading KaTeX, which doesn't work in quirks mode. One case for including the change in this PR is that #212 is one of this PR's motivations.

srid: Okay, feel free to leave it in then - but do add that note (re: KaTeX) as a code comment above that.

srid (approved):

vizziv:
Okay, I undid the config changes (except I removed the `mathJaxSupport` field) and explained DOCTYPE.

The `Neuron.Config` and `Neuron.Config.Type` commit histories on this branch are now kind of messy, so squashing these commits upon merging makes sense.

PR #395, opened 2020-09-13
Add missing modules in neuron.cabal

Lysxia:


Lysxia:
Oh wait no, there is no HeadHTML

srid:
HeadHtml should really be in src/app, not src/lib - because library cannot and should not depend on `shake` (it must compile on ghcjs).

Lysxia:
Okay, should be fixed

Lysxia:
Oh lol I failed my fix. Thanks!

PR #396, opened 2020-09-14
Add lower bound dependent-sum >= 0.7

Lysxia:
Makes available foldSome which is used in neuron/src/app/Neuron/CLI/Open.hs


srid:
You'll want to put this on `app-common` rather than `library-common`. CLI/Open.hs uses the former. And the later should allow a lesser version for reflex-platform compat (which is why there is no version bound).

PR #403, opened 2020-09-15
Create private cabal library shared by executable and test suite

### LINKED ISSUE #402: Redundant build
When you install neuron, the library gets built 3 times:

- Once by the library
- Once by the executable
- Once by the test suite

This is because both the executable and the test suite mention `src/lib` in `hs-src-dirs`. They should not. Similarly, the test suite should not use the executable's `src/app`. If two or more components need to depend on the same thing, it should be put in a library; this seems a good use case for a private library in `neuron.cabal`.

Lysxia:
Fixes #402


Lysxia:
This should keep the ghcid trick now, though I haven't tested; it will be a few days before I get around to test it. Basically, if the `ghcid` flag is set, we don't import the neuron libraries and just make the whole `src` directories visible to every cabal component.

srid:
Note to self: will have to confirm that ghcide/HLS performance (startup time) is not impacted adversely.

srid:
@Lysxia Thanks for the PR; I've merged it in #413

Issue #183, opened 2020-05-14
Static binaries for releases


itamarst:
As someone else commented on lobste.rs, it would be nice to have pre-generated static binaries so users who don't have Nix and Haskell setup don't have to install a whole new toolchain to use Neuron.

---

[text below added by @srid]

- [x] Static binaries for Linux (and Windows, through WSL2)
- [ ] Static binaries for macOS
- [x] Host static binaries for release, and link to from docs

srid:
I could use some help on this. It seems we will need to use this: https://github.com/nh2/static-haskell-nix#static-haskell-nix

itamarst:
FWIW a semi-static executable that only depends on libc would probably be fine for most people? Unless you're linking against some other C libraries.

srid:
FWIW, this is what the current executable links to (on linux):

```sh
$ ldd (which neuron)
        linux-vdso.so.1 (0x00007ffe79d92000)
        libm.so.6 => /nix/store/pnd2kl27sag76h23wa5kl95a76n3k9i3-glibc-2.27/lib/libm.so.6 (0x00007f7bb2490000)
        libz.so.1 => /nix/store/3bfzi6w3067hbdyl15vmypjn4q6az9j6-zlib-1.2.11/lib/libz.so.1 (0x00007f7bb2471000)
        liblua.so.5.3 => /nix/store/yaim51rzjyrp2dia8d954fywnwkp77na-lua-5.3.5/lib/liblua.so.5.3 (0x00007f7bb2436000)
        libncursesw.so.6 => /nix/store/zp8v7wkbplvmzyds87icn1rk070hizdc-ncurses-6.1-20190112/lib/libncursesw.so.6 (0x00007f7bb23c4000)
        libpthread.so.0 => /nix/store/pnd2kl27sag76h23wa5kl95a76n3k9i3-glibc-2.27/lib/libpthread.so.0 (0x00007f7bb23a3000)
        librt.so.1 => /nix/store/pnd2kl27sag76h23wa5kl95a76n3k9i3-glibc-2.27/lib/librt.so.1 (0x00007f7bb2397000)
        libutil.so.1 => /nix/store/pnd2kl27sag76h23wa5kl95a76n3k9i3-glibc-2.27/lib/libutil.so.1 (0x00007f7bb2392000)
[...]

PR #417, opened 2020-09-29
Build static binaries on Linux

srid:
For #183 

- [x] Build and test on Linux
- [ ] Build and test on macOS


srid:
`pkgsMusl` works on Linux, but unsupported on macOS.

`pkgsStatic` is supported on both platforms, but `lua` (which `pandoc` depends on) doesn't build on it:

image

srid:
Notes https://www.srid.ca/db359075.html

Issue #275, opened 2020-07-06
Neuron search, Org format not recognized


Whil-:
Hi again =)

Trying out the Neuron CLI. And it seems `neuron search` won't recognize files in Org format.

Zettelkasten folder content:
```
gustav@Kontor:~/zettelkasten$ ls -l
total 0
-rw-rw-rw- 1 gustav gustav  96 Jul  6 21:17 06ae322c.md
-rw-rw-rw- 1 gustav gustav  92 Jul  6 21:55 8a15ec9a.org
-rw-rw-rw- 1 gustav gustav  70 Jul  6 21:17 a29e0fde.md
-rw-rw-rw- 1 gustav gustav 145 Jul  6 21:45 neuron.dhall
```

And `neuron search` will only list the two markdown files. Same result with `--full-text` added to the command.

srid:
This script https://github.com/srid/neuron/blob/master/neuron/src-bash/neuron-search should be made to take the `*.md` as an argument.

clojj:
trying to test this.. got this problem:
running `bin/run -d $(pwd)/guide search -a` gives me this error
![grafik](https://user-images.githubusercontent.com/4367963/90392335-f34b5600-e08e-11ea-9c84-7cd3383061cd.png)

clojj:
in a clean branch, I could not start bin/run ... search 
maybe because of the bash script, this command can only be tested with a neuron executable?

srid:
I don't see that compile error; in any case, you don't really want to use bin/run to run interactive commands like `search`, because ghcid doesn't seem to handle that well. Run `nix-build` and test `./result/bin/neuron`.

srid:
Closing, and moving what's left to #422

PR #420, opened 2020-10-01
Make .org recognized by Neuron search

pt2121:
Addresses #275


neuron/src/app/Neuron/CLI/Search.hs
```diff
@@ -23,8 +23,8 @@ searchScriptArgs :: SearchCommand -> [String]
 searchScriptArgs SearchCommand {..} =
   let searchByArgs =
         case searchBy of
-          SearchByTitle -> ["(^# )|(^title: )", "2"]
-          SearchByContent -> ["", "2"]
+          SearchByTitle -> ["(^# )|(^title: )", "2", "/*.{md,org}"]
+          SearchByContent -> ["", "2", "/*.{md,org}"]
```
srid, line 27: Instead of hardcoding this value (`"/*.{md,org}"`), we should use [`getZettelFormats`](https://github.com/srid/neuron/blob/8d46c5829297ac5a0030b712d9fc74819198b39f/neuron/src/lib/Neuron/Config/Type.hs#L45) to determine it automatically. You can pass the config value to `interactiveSearch` and use it here.

So that if someone has org files lying around, and they are not enabled in neuron.dhall, then `neuron search` should not be searching those. And vice-versa for Markdown files.

srid (changes requested):
Thanks for the PR; just one thing to address.

neuron/src/app/Neuron/CLI/Search.hs
pt2121: Thanks, will update.

neuron/src/app/Neuron/CLI/Search.hs
```diff
@@ -23,8 +23,8 @@ searchScriptArgs :: SearchCommand -> [String]
 searchScriptArgs SearchCommand {..} =
   let searchByArgs =
         case searchBy of
-          SearchByTitle -> ["(^# )|(^title: )", "2"]
-          SearchByContent -> ["", "2"]
+          SearchByTitle -> ["(^# )|(^title: )", "2", "/*.{md,org}"]
```
srid, line 26: By the way, another issue: the `^#` here won't work for org files, which use `*` prefix for the title. You will want to test the change on actual org files with specific titles, and make sure that searching for the title matches them.

pt2121: Sorry about that I'm new to the `.org`. I'll look into it.

neuron/src/app/Neuron/CLI/Search.hs
```diff
@@ -11,27 +11,37 @@ module Neuron.CLI.Search
...
-searchScriptArgs SearchCommand {..} =
-  let searchByArgs =
+searchScriptArgs :: (NonEmpty ZettelFormat) -> SearchCommand -> [String]
+searchScriptArgs formats SearchCommand {..} =
+  let extensionPattern = "/*{" <> (Text.unpack $ Text.intercalate "," $ toList $ zettelFormatToExtension <$> formats) <> "}"
+      searchByArgs =
         case searchBy of
-          SearchByTitle -> ["(^# )|(^title: )", "2"]
-          SearchByContent -> ["", "2"]
+          SearchByTitle -> ["(^# )|(^title: )|(^\\* )", "2", extensionPattern]
```
pt2121, line 34: I haven't found a good way to make a pattern specific to a correct extension. Any suggestions?

neuron/src/app/Neuron/CLI/Search.hs
pt2121: Done!

neuron/src/app/Neuron/CLI/Search.hs
srid: Hmm, that's unfortunate. 

If no solution can be found, we should probably remove the `^\\*` matching entirely and just say in the documentation that `neuron search` (not full-text) is unsupported for .org files. Your PR would then effectively make it work for full-text search only.

srid: In that `neuron search` should fail right away, if org is enabled in neuron.dhall. But `neuron search --full-text` should work fine regardless. 

Unless you have got other suggestions ...

pt2121: Thanks for the pointer. I'll look into it.

pt2121: Updated. Can you review when having a chance?

srid (approved):
Looks good.

Issue #437, opened 2020-10-16
Raw HTML is not interpreted as one


flyinggrizzly:
Inserting the following HTML into a note:

```markdown
<!-- inline to prevent possible interpretation of indentation as code block -->
<video><source src='./static/otaku-builder-sdcs-unicorn-destroy-mode.mp4' type='video/mp4' /></video>
```
results in the following in the generated page:

![image](https://user-images.githubusercontent.com/10967676/96245491-46faf280-0f9f-11eb-842a-42a4ca425974.png)

```html
<p>
  <a href="video">video</a>
  <span class="ui left pointing red basic label">links to non-existant zettel: video</span>
  <source src="./static/otaku-builder-sdcs-unicorn-destroy-mode.mp4" type="video/mp4">
  <a href="/video">/video</a>
</p>
```

I can embed the video with an iframe, but I prefer to actually save files that are potentially ephemeral (like Youtube videos).

srid:
Raw HTML needs to be wrapped in code-blocks, in order to disambiguate the meaning of "<video>" (how can neuron tell it is not a link to `video.md`? Try:

```markdown
    ``` {=html}
    <video><source src='./static/otaku-builder-sdcs-unicorn-destroy-mode.mp4' type='video/mp4' /></video>
    ```
```

Example: https://raw.githubusercontent.com/srid/guide.cerveau.app/master/index.md

srid:
Documentation on this custom syntax: https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/raw_attribute.md

flyinggrizzly:
Yup, that works. I’m also now seeing the autolink footnote in the Neuron guide, which I’m assuming why Neuron has the issue with (some, but not all?) HTML elements. 

Given my expectation that valid HTML is valid markdown, this definitely took me by surprise.

(I’m trying to figure out how to phrase this that’s not just entitled user, but I’m sorry if I miss the mark)—would it be worth considering deprecating the auto link behavior in Neuron now that wiki links are recommended, so that there isn’t the issue with HTML in MD? Not fully supporting HTML in the markdown is definitely a surprise for people who are used to MD, especially anyone who uses it with existing static site generators. 

This would make migrating potentially easier for folks coming from pure MD environments. 

But I’m also very aware that it would leave existing Neuron users with a problem...

If that breaking change isn’t something you want to do, maybe something about this in the Zettel Markdown docs at least would help others who run into the same issue?

[...]

PR #442, opened 2020-10-19
Documentation for using raw HTML in notes

flyinggrizzly:
This PR adds documentation for using HTML in Neuron markdown.

Because Neuron scans notes for links before passing them to the MD parser, and because Neuron still supports `<note-name>` link syntax, it treats simple HTML elements as notes, which result in missing/unknown note warnings.

The documentation added covers two solutions:

1. Using Commonmark raw-attribute syntax as noted in https://github.com/srid/neuron/issues/437#issuecomment-710698849
2. Adding an HTML attribute to the element

Adds documentation for #437

---

## Consideration

I discovered the option of adding HTML attrs when I was wondering why an `iframe` worked in a note without the raw-attribute syntax, but `video` didn't. It has worked for every basic HTML element I've tried to use it for without issue.

The upside, is that the Markdown is more idiomatic, and less tied to Commonmark. The downside is the proliferation of classes/attributes. However, the attribute only appears to be required on the outmost HTML element of a block.

### DIFF
--- a/guide/markdown.md
+++ b/guide/markdown.md
@@ -2,7 +2,6 @@
 
 Zettel files are written in Markdown[^other], per the [CommonMark](https://commonmark.org/) specification. Neuron uses [commonmark-hs](https://github.com/jgm/commonmark-hs) to parse them into the [Pandoc AST](https://pandoc.org/using-the-pandoc-api.html), as well as provides an extention on top to handle zettel links.
 
-* [[[using-html-in-markdown]]]
 * [[[linking]]]
 * [[[tags]]]
 * [[[metadata]]]
@@ -10,5 +9,6 @@ Zettel files are written in Markdown[^other], per the [CommonMark](https://commo
 * [[[math]]]
 * [[[2016401]]]
 * Styling elements using Semantic UI ([\#176](https://github.com/srid/neuron/issues/176))
+* [[[raw-html]]]
 
 [^other]: Neuron is designed to be extended with other markup formats as well. Org-mode is currently supported (see the `formats` setting in [[configuration]]) but not all features work with it. Neuron recommends Markdown, which is supported everywhere including [[cerveau]].  
--- a/guide/markdown.md
+++ b/guide/markdown.md
@@ -2,6 +2,7 @@
 
 Zettel files are written in Markdown[^other], per the [CommonMark](https://commonmark.org/) specification. Neuron uses [commonmark-hs](https://github.com/jgm/commonmark-hs) to parse them into the [Pandoc AST](https://pandoc.org/using-the-pandoc-api.html), as well as provides an extention on top to handle zettel links.
 
+* [[[using-html-in-markdown]]]
 * [[[linking]]]
 * [[[tags]]]
 * [[[metadata]]]
--- a/guide/using-html-in-markdown.md
+++ b/guide/using-html-in-markdown.md
@@ -20,7 +20,7 @@ From the docs:
 
 > If attached to a fenced code block, it causes the block to be interpreted as raw block content with the specified format.
 
-For HTML, this just requires adding `{=html}` after the opening ``` codeblock fence.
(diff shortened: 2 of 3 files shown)


guide/using-html-in-markdown.md
```diff
@@ -0,0 +1,83 @@
+# Using HTML in Markdown, in Neuron
+
+Usually, Markdown supports raw HTML inline as valid Markdown syntax.
```
srid, line 3: Since neuron uses `commonmark` (which does support raw HTML), you should link to the spec: https://spec.commonmark.org/0.28/#raw-html

guide/using-html-in-markdown.md
```diff
@@ -0,0 +1,83 @@
...
+This may be changed in the future if a reliable migration path is identified (i.e. a script that would automatically convert angle-bracket links to wiki-links).
+
+## Getting HTML to work
+
+In order to get raw HTML to work in a note, for example to embed a `<video />` element, there are two options:
+
+1. Wrap the HTML in a code-block with a 'raw-attribute', which causes the code block to be interpreted as raw inline content
+2. Add an HTML attribute to your element, like `class` or `name`, which causes Neuron to recognize that element is not a note link
+
+### Using raw-attributes to insert HTML
```
srid, line 15: Better to title it as "Using fenced code blocks"

guide/using-html-in-markdown.md
```diff
@@ -0,0 +1,83 @@
...
+
+results in the following generated HTML:
+
+```````````````````````````````` example
+<h2>Some markdown interspersed with HTML</h3>
+<p>Here is a video:</p>
+<video><source src='static/video.mp4 /></video>
+````````````````````````````````
+
+### Using HTML attributes to cause Neuron to recognize HTML elements
```
srid, line 45: Better: "Using HTML attributes"

guide/using-html-in-markdown.md
```diff
@@ -0,0 +1,83 @@
...
+
+If an HTML attribute is added to an element, Neuron ignores it while generating the link heterarchy and passes it unchanged to the Markdown parser.
```
srid, line 47: Replace with: "If the top-level HTML element has one or more attributes, neuron will never parse it as a link".

guide/using-html-in-markdown.md
```diff
@@ -0,0 +1,83 @@
...
+
+This will likely happen without notice for elements like `iframe` (say, a Youtube embed), since they depend heavily attribute values. For simpler elements, something as small as adding an innocuous and arbitrary class works:
+
+
+```````````````````````````````` example
+## Some markdown interspersed with HTML
+
+Here is a video:
+
+<video class="force-html"><source src='static/video.mp4 /></video>
```
srid, line 57: I suppose you can instead add `id=some-video`, which is probably better (to avoid using some accidental CSS class).

srid (changes requested):
Awesome, thanks for the elaborate documentation on the topic! Just a few things to address before the merge.

guide/markdown.md
```diff
@@ -2,6 +2,7 @@
 
 Zettel files are written in Markdown[^other], per the [CommonMark](https://commonmark.org/) specification. Neuron uses [commonmark-hs](https://github.com/jgm/commonmark-hs) to parse them into the [Pandoc AST](https://pandoc.org/using-the-pandoc-api.html), as well as provides an extention on top to handle zettel links.
 
+* [[[using-html-in-markdown]]]
```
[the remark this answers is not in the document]
srid: And should also be added to the list, at the bottom (not front).

flyinggrizzly:
I've added 3 commits that should resolve the requested changes. Let me know if you'd like me to rebase this down to one or two commits before merging!

srid:
Thanks! This had slipped out of my radar accidentally; merged now.

flyinggrizzly:
No worries--I'm glad I could help!

PR #443, opened 2020-10-19
Exclude punctuation from inline tag parse

### LINKED ISSUE #441: exclude punctuation marks from inline tags
Inline tags are convenient compared with metadata, but the current parsing method is not intuitive.

For example, I would naturally write something like this:

```markdown
.....It seems that the best tool I can find is #Zotero, which is very helpful for academic #writing.
```

And the comma and period is included in the tags, which is not intended.

Another example is that when exported from some other tools, the note is very likely to list tags like this:

```markdown
# An introduction to language

## Metadata

* Type: Book
* Authors: Victoria Fromkin, Robert Rodman, Nina M. Hyams
* Date: 2014
* Date added: 2020-08-15
* Cite key: fromkinIntroductionLanguage2014
* Related: 
* Tags: #Zotero, #ToRead, #Language and Languages, #Linguistics

```

wbadart:
Resolves #441

I was thinking about something a little more extensible, e.g.

```haskell
some (noneOfToks $ [Spaces, UnicodeSpace, LineEnd] <> fmap Symbol excludedChars)
where excludedChars = "],."
```

but I figured this list probably won't be growing or changing much, and I figured the simple/ expanded version I'm proposing reads a little more smoothly. Let me know if you have a preference one way or the other! (Also please let me know if any other characters should be excluded from the parse.)

### DIFF
--- a/neuron/src/lib/Neuron/Reader/Markdown.hs
+++ b/neuron/src/lib/Neuron/Reader/Markdown.hs
@@ -223,7 +223,9 @@ wikiLinkSpec =
 
 inlineTagP :: Monad m => P.ParsecT [CM.Tok] s m [CM.Tok]
 inlineTagP =
-  some (noneOfToks [Symbol ']', Symbol ',', Symbol '.', Spaces, UnicodeSpace, LineEnd])
+  some (noneOfToks $ [Spaces, UnicodeSpace, LineEnd] <> fmap Symbol punctuation)
+  where
+    punctuation = "[];:,.?!"
 
 -- rawHtmlSpec eats angle bracket links as html tags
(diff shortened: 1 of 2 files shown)


srid (changes requested):
Thanks! 

Could you add tests [here](https://github.com/srid/neuron/blob/e7568ca5f51609bb406a48527b5ba52d31d11f9c/neuron/test/Neuron/Zettelkasten/Zettel/ParserSpec.hs#L23-L37)?

I prefer a `where punctuationChars = "],."`, because it makes it very clear just by reading the code as to why these characters are being excluded. 

Also, why add only comma and dot? What about other punctuation letters like semi-colon, colon? What about `[`? We should probably add them as long as the tests don't fail.

wbadart:
I only included the comma and period on my first pass because those were the only characters tripping my personal workflow :)

I'll work in a more complete set with my next commit!

PR #445, opened 2020-10-22
make search include zettel with no explicit title, use filename instead

LATBauerdick:
I "hacked" `neuron/src-bash/neuron-search` to include all zettel, also those that previously were not listed because they don't have an explicit `title: ` in the header, nor a `# <title>` line. 

I have many "legacy" zettel imported from iA writer, Ulysses, nvAlt and such where the convention was to use `## <title>` to generate filename etc. Currently they don't show up in the `neuron search` function. That's really not good. 

I think the `neuron` convention is to use the filename as title if there is none specified, which this PR takes care of.

It's a bit of a kludge, but it seems to work fine. Maybe someone with more `bash` expertise can clean it up later. 

Also changed the `xarg` call to deal with filenames that contain blanks.

### DIFF
--- a/neuron/src-bash/neuron-search
+++ b/neuron/src-bash/neuron-search
@@ -8,6 +8,8 @@ EXTENSIONS=${4}
 OPENCMD=$(envsubst -no-unset -no-empty <<<${5})
 cd ${NOTESDIR}
 
+# The LC_ALL here is fix the "illegal byte sequence" error
+# See https://github.com/srid/neuron/pull/445#discussion_r513066111
 {    rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
   && rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
      --files-without-match | awk '{ printf "%s:!title: %s\n", $0, $0}'; } \
--- a/neuron/src-bash/neuron-search
+++ b/neuron/src-bash/neuron-search
@@ -8,10 +8,13 @@ EXTENSIONS=${4}
 OPENCMD=$(envsubst -no-unset -no-empty <<<${5})
 cd ${NOTESDIR}
 
-{ ls *.md | awk '{ printf "%s:!title: %s\n", $0, $0}' && rg --no-heading --no-line-number --with-filename "${FILTERBY}" -g "${EXTENSIONS}"; } | sort -t: -k1,2 -r | sort -t: -k1,1 -u \
+{    rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
+  && rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
+     --files-without-match | awk '{ printf "%s:!title: %s\n", $0, $0}'; } \
+  | ( [ ${FILTERBY} ] && LC_ALL=C sort -t: -k1,2 -r || cat ) \
+  | ( [ ${FILTERBY} ] && LC_ALL=C sort -t: -k1,1 -u || cat )\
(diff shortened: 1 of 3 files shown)


LATBauerdick:
ok, there's a problem that the bin/test didn't catch, working on it... (search -a flag is seemingly broken)

srid:
@LATBauerdick Is this PR still in work-in-progress?

srid:
CC @jared-w as he had some ideas.

LATBauerdick:
I committed an update that works just fine. I'd be happy to include other ideas, though, @jared-w please let me know!

jared-w:
What you did was essentially the solution I came up with, actually. I had some additional over-engineering that I had thought of in order to support arbitrary file extension support with different regexes (eg a regex for matching org mode titles, and one for matching markdown titles)

The idea was essentially to use `IFS` and split on newlines to pass more structured data as args into the script

```bash
IFS=$'\n' read -r -d '' -a exts <<< "${EXTENSIONS}"
IFS=$'\n' read -r -d '' -a fltrs <<< "${FILTERBY}"

declare -a titles
for i in "${!exts[@]}"; do
  mapfile -t "${#titles[@]}" titles < <(rg --files-without-match --no-heading --no-line-number --with-filename --sort path "${fltrs[$i]}" -g "*.${exts[$i]}" | awk -F '.' '{ORS="";print $0":";NF--;print $0;print "\n"}')
  mapfile -t "${#titles[@]}" titles < <(rg --no-heading --no-line-number --with-filename --sort path "${fltrs[$i]}" -g "*.${exts[$i]}")
done

printf '%s\n' "${titles[@]}" | \
  fzf ...
```

Something like this.

neuron/src-bash/neuron-search
```diff
@@ -7,8 +7,14 @@ SEARCHFROMFIELD=${3}
 EXTENSIONS=${4}
 OPENCMD=$(envsubst -no-unset -no-empty <<<${5})
 cd ${NOTESDIR}
-rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
+
+{    rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
+  && rg --no-heading --no-line-number --with-filename --sort path "${FILTERBY}" -g "${EXTENSIONS}" \
+     --files-without-match | awk '{ printf "%s:!title: %s\n", $0, $0}'; } \
+  | ( [ ${FILTERBY} ] && LC_ALL=C sort -t: -k1,2 -r || cat ) \
```
srid, line 14: Why is `LC_ALL` being set here? Is that necessary for this PR?

srid: Also, what does the `sort` do here? A search for `link` shows: 

image

which appears to be the inverse of "select the most matching item" (which is 'Uplink Tree' IIUC).

srid: Okay, that sort order is the same as `master`, so I guess that's irrelevant. Also, the original command had a `--sort path`, which I think sorts by filename.

srid:
@LATBauerdick I've tested this locally, and it looks good. Except for the LC_ALL thing (I'm not sure what it is for), it is good for a merge.

neuron/src-bash/neuron-search
LATBauerdick: I ran into a problem with `sort` when it complained about "illegal byte sequence". Most likely one of the files was indeed "bad", but the failure mode was ugly in that a normal user would have no idea what's going on, and the `LC_ALL` fixed that. If there is a way to point out what file has the "illegal byte sequence" the user could just remove or fix that file, but error reporting in this context looks hard, so I thought it better to just avoid the problem. 

But I have no strong opinion on leaving it in or removing it.

srid:
Since this works without breaking existing behaviour (as far as I can verify), let's merge it. But soon enough we should discard all these bash magic, and just do things in Haskell (while delegating to fzf and bat wherever appropriate).

Issue #439, opened 2020-10-16
Zettels with title ID, but no header, are unavailable in `neuron search`


srid:
The [search script](https://github.com/srid/neuron/blob/master/neuron/src-bash/neuron-search) should be adapted to deal with implicit title in the filename, when one is not provided in the note body. 

To reproduce:

```
echo Test > "Foo bar.md"
neuron search
```

"Foo bar" won't appear in search.

srid:
Either extend the `neuron-search` script (like #445), or rewrite the bulk of it in Haskell.

LATBauerdick:
this works now with the merge of [445](https://github.com/srid/neuron/pull/445). Could be closed, although the Haskell rewrite is still desirable.

srid:
@LATBauerdick This is *almost* fixed by #445, but it doesn't handle an edge case, wherein *all* of your notes use title ID.

To repro:

```
mkdir /tmp/repro
cd /tmp/repro/ 
echo Test > "Foo bar.md"
echo Test2 > "Qux.md"
neuron search
```

LATBauerdick:
hmm, embarrassing mistake... need to replace the `&&` in `neuron/search-bash/` with a `;`, that's it...

PR #455, opened 2020-11-03
Update neuron-search

LATBauerdick:
fix for search not returning anything in case none of the zettle has an explicit title

### DIFF
--- a/neuron/src-bash/neuron-search
+++ b/neuron/src-bash/neuron-search
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -uo pipefail
 NOTESDIR=${1}
 FILTERBY=${2}
 SEARCHFROMFIELD=${3}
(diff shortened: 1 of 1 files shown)


srid:
@LATBauerdick Have you tested this change using the repro from https://github.com/srid/neuron/issues/439#issuecomment-721405536 ? It doesn't seem to fix it.

LATBauerdick:
> @LATBauerdick Have you tested this change using the repro from [#439 (comment)](https://github.com/srid/neuron/issues/439#issuecomment-721405536) ? It doesn't seem to fix it.

my latest commit should do it -- I'm starting to wonder if `bash` programming is harder than `Haskell`?

srid:
Instead of suppressing errors in all commands (by removing `set -e`), it would be better to suppress errors is just that one command that should be ignored, by adding `|| echo`. Fixed in da44df0

Yes, bash has always been annoying to use, which is another reason we should move bulk of it to Haskell. At minimum, we can basically use `neuron query --graph` to get a list of all filenames along with their titles, and then construct a CLI that uses fzf, passing these filenames and titles, before executing it.

PR #470, opened 2020-11-16
Fix more parsing issues with pipe characters

### LINKED ISSUE #465: `|` breaks lists
If `|` ist placed somewhere in a list, it is not rendered correctly but instead produces a paragraph in HTML:

```markdown
- a
- |
- b
```
```html
<div class="pandoc">
  <ul>
    <li>a</li>
  </ul>
  <p>- |</p>
  <ul>
    <li>b</li>
  </ul>
</div>
```
I noticed this because I have a list containing mathematical expressions, such as `$|x|$`. If I run pandoc directly to output HTML it works as intended.

### LINKED ISSUE #469: `|` breaks headers
version: neuron-1.1.6.2

I have a lot of files that have a pipe in the header title of the note

this example breaks the header
```
---
date: 2020-11-16T06:39
tags: [anarchismus,capitalism,collapse,date/2020/11/16,franco-berardi,mind/history]
---

# Extinction, Rebellion, Happiness | SchwarzerPfeil by schwarzerpfeil.de
```

to fix this I need to escape the pipe

```
---
date: 2020-11-16T06:39
tags: [anarchismus,capitalism,collapse,date/2020/11/16,franco-berardi,mind/history]
---

# Extinction, Rebellion, Happiness \| SchwarzerPfeil by schwarzerpfeil.de
```

without the fix the filename is displayed 

> \# 1d0779db7445ff68b1f70c57bb7c9146-Extinction,_Rebellion,_Happiness_SchwarzerPfeil_by_schwarzerpfeil.de.md

eigenzeugs:
`pipeTableSpec` should also be after `defaultSyntaxSpec`

Fixes #469 & fixes #465

### DIFF
--- a/neuron/src/lib/Neuron/Reader/Markdown.hs
+++ b/neuron/src/lib/Neuron/Reader/Markdown.hs
@@ -130,13 +130,15 @@ neuronSpec =
       CE.rawAttributeSpec,
       CE.fencedDivSpec,
       CE.bracketedSpanSpec,
-      CM.defaultSyntaxSpec
+      CM.defaultSyntaxSpec,
+      -- as the commonmark documentation states, pipeTableSpec should be placed after
+      -- fancyListSpec and defaultSyntaxSpec to avoid bad results when non-table lines
+      CE.pipeTableSpec
     ]
   where
     -- Emoji extension introduces ghcjs linker issues
     gfmExtensionsSansEmoji =
       CE.strikethroughSpec
-        <> CE.pipeTableSpec
         <> CE.autoIdentifiersSpec
         <> CE.taskListSpec
 


srid:
@eigenzeugs I had to revert your last PR, because it broke some GFM functionality. See #472

I suppose a fix would be to relocate only `pipeTableSpec`, and not the entirety of `gfmExtensionsSansEmoji`.

In any case, we should ideally add unit tests to catch regression like this. Or, do a manual testing to make sure that existing Markdown features (those of the commonmark extensions in use) continue to work.

eigenzeugs:
Oh I'm sorry, should have tested more. Moving only `pipeTableSpec` seems to work as you suggested. I manually tested the other extensions this time, but if this requires more testing please let me know. How would you prefer a unit test for this to look like?

Issue #480, opened 2020-11-26
Org-mode folgezettel / uplinks fail to work


beandipper:
After testing the org format within the guide zettelkasten using the nix-shell --pure environment, I was able to determine that that the linking doesn't work as expected.

  What happened was that folgezettel links were treated as normal links, as if they had the ?cf postfix.

 To test this I created two org zettels. 
1. The first one is a copy of the org.md in the guide convrted to org mode org.md ~> orgTest.org, including tags and links to .md zetteln
2. The second one, linker.org is just a simple org zettel that contains a single folgezettel link to the one created in the previous step 

```
* Linker Org Zettel

  - A folgezettel link with label [[z:/orgTest][org test title]]
  - A folgezettel link without label [[z:/orgTest]]
  - A link without prefix [[orgTest?cf]].
  - A folgezettel link without prefix [[orgTest]].
```

- From the above links none of the folgezettel links are detected as so
- Link labels work whether specified with custom labels, or automatically detected, as in the second link displayed in HTML with the orgTest.org title
- Links without the z: prefix aren't detected as they are in markdown mode
- Links with other parameters such as tags aren't being tested yet. This can be done if wished

The JSON for the orgTest.org in question doesn't imply that folgezettel connections are working. 

A snippet showing only partial results for one folgezettel link from: 
```
./bin/run -d $(pwd)/guide query --uplinks-of orgTest
```
is
```
{
  "skipped": {},
  "result": [
    [
      [
        "Folgezettel",
        [
          {
            "t": "Plain",
            "c": [
              {
                "t": "Str",
                "c": "Linking by tag: "
              },
              {
                "t": "Code",
                "c": [
                  [
                    "",
                    [],
                    []
                  ],
                  "experimental"
                ]
              }
            ]
          }
        ]
      ],
      {
        "zettelTags": [],
        "zettelDate": null,
        "zettelID": "experimental",
        "zettelError": null,
        "zettelUnlisted": false,
        "zettelContent": [],
        "zettelQueries": [
          [
            [
              "ZettelQuery_ZettelsByTag",
              [
                [
                  "experimental"
                ],
                "Folgezettel",
                {
                  "zettelsViewLinkView": "LinkView_Default",
                  "zettelsViewGroupByTag": false
                }
              ]
            ],
            [
              {
                "t": "Plain",
                "c": [
                  {
                    "t": "Str",
                    "c": "Linking by tag: "
                  },
                  {
                    "t": "Code",
                    "c": [
                      [
[...]

PR #481, opened 2020-11-26
Fix for org-mode folgezettel links

beandipper:
As described in #480 , I wasn't able to get folgezettel links to work with org-mode formatted Zetteln


beandipper:
After talking with @srid, he decided that org-mode should use the following formats for links, without the need for this PR
1. standard non-folgezettel link : [[z:/foo]]
2. folgezettel-link: [[z:/foo?type=branch]]

PR #490, opened 2020-12-06
Advance nixpkgs (GHC 8.10)

### LINKED ISSUE #474: base upper bound is incompatible with ghc 8.10
thx!

srid:
Resolves #474

- [x] Test that nothing is broken on ghc 8.10
- [ ] Advance nixpkgs further to use https://github.com/NixOS/nixpkgs/commit/15b3bc33064eeb0cba743cad585c829b6694669c for static.nix and build static binary


madjar:
I tested it on my setup (I try to follow nixpkgs-unstable), and all seems find: my generated neuron website looks good, and the `query` and `search` command look fine.

As a side note, it seems that the fix for the footnote bug in commonmark has been released, and that the newest release is in the nixpkgs you've advanced to. So it should be possible to stop using commonmark from git.

srid:
It looks like we might just have to drop support for static binary builds. That's not building with the newer nixpkgs, and I don't desire to maintain this for eternity.

PR #512, opened 2021-01-01
 Clarify that directories/note IDs must be globally unique when using the dirtree plugin

flyinggrizzly:
This is the only thing I stumbed on--otherwise the documentation for the `dirtree` plugin was clear for me.

### DIFF
--- a/guide/Plugins/Directory Tree.md
+++ b/guide/Plugins/Directory Tree.md
@@ -10,7 +10,9 @@ The *Directory Tree* plugin automatically creates a [[folgezettel-heterarchy]] r
 
 ## Directory Zettel
 
-Given a folder `./Project/HouseWarming/`, this plugin will create two zettels with [[id]] `Project` and `HouseWarming`. Directory names must be unique, inasmuch they directly reflect [[id]] which must also be unique. In the [[web]], neuron will display the "contents" of a directory beneath the zettel content. 
+Given a folder `./Project/HouseWarming/`, this plugin will create two zettels with [[id]] `Project` and `HouseWarming`. In the [[web]], neuron will display the "contents" of a directory beneath the zettel content.
+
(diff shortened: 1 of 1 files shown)


guide/Plugins/Directory Tree.md
```diff
@@ -10,12 +10,14 @@ The *Directory Tree* plugin automatically creates a [[folgezettel-heterarchy]] r
...
 
 The zettel content for directory zettels are by default empty. However, you may explicitly specify them by creating zettel files using the same [[id]]; for the example above, you may create `./Project.md` and `./Project/HouseWarming.md` - and they will be "merged" to the auto-created directory zettels.
 
 ## Links & tags created
 
 A [[folgezettel-heterarchy]] is formed, reflecting the directory tree on disk, with the exception of the `index` zettel[^man]. The forming of links is based on hierarchical [[tags]] (reflecting the relative path), that every normal zettel and directory zettel is automatically tagged with.
 
-[^man]: You may manually form these relationships by adding `[[[z:zettels?tag=root]]]` to the top-level `index.md`
\ No newline at end of file
+[^man]: You may manually form these relationships by adding `[[[z:zettels?tag=root]]]` to the top-level `index.md`
```
flyinggrizzly, line 23: This change appears to be just a whitespace change caused by saving in the GH web editor.

srid:
Thanks. Will merge.

We should further explain the rationale and utility behind this uniqueness criteria.  In `work/notes/`, "notes" is a bad name for an [atomic](https://neuron.zettel.page/atomic.html) zettel (usually it would be someething like `work/Project1/` or `work/Potential clients/`). 

A good example would be `2020/12/2020-12-02` - which would conflict with other years' December, and can be solved by renaming the folder to "2020/2020-12/2020-12-02". The benefit is ... simplicity in linking, without complicated linking strategies in neuron, spamming user's mental model. Just link to year, month or day using their file/directory name.  It allows you to have a `2020-20.md` zettel, which gets "merged" automatically with the `2020-20/` directory zettel, and you can link to it as `[[2020-20]]` without banging the head against the wall thinking in terms of filesystem paths.

flyinggrizzly:
I'll see if I can't get something concise and useful drafted about ID atomicity.

PR #517, opened 2021-01-03
Add label to the Home link

### LINKED ISSUE #342: Make the index link easily recognizable
From an user:

> I couldn’t find an index button then discovered it was the 🌲. Might it be an idea to have something more recognisable?

Can be done by making it a [labelled icon](https://fomantic-ui.com/collections/menu.html#labeled-icon).

rvl:
Resolves #342

This adds a "Home" label to the menu item.

Perhaps labels on all of the menu items? Is the "compact" style alright? I think it's OK as it is though.

Perhaps the Home item could have an icon in the impulse frontend? I could do that.

### Screenshot

1. Labelled home button and DOM

Screenshot from 2021-01-04 07-41-17

2. Illustration of label without space

home-space

### DIFF
--- a/neuron/src/lib/Neuron/Frontend/Zettel/View.hs
+++ b/neuron/src/lib/Neuron/Frontend/Zettel/View.hs
@@ -151,8 +151,9 @@ renderBottomMenu themeDyn mIndexZettel mEditUrl = do
       ffor x $ \case
         Nothing -> blank
         Just indexZettel -> do
-          neuronDynRouteLink (Some . Route_Zettel . Z.zettelSlug <$> indexZettel) ("class" =: "item" <> "title" =: "Home") $
+          neuronDynRouteLink (Some . Route_Zettel . Z.zettelSlug <$> indexZettel) ("class" =: "item" <> "title" =: "Home") $ do
             semanticIcon "home"
+            text " Home"
     -- Edit url
     forM_ mEditUrl $ \editUrl -> do
       let attrs = "href" =: editUrl <> "title" =: "Edit this page"


neuron/src/lib/Neuron/Frontend/Zettel/View.hs
```diff
@@ -151,8 +151,9 @@ renderBottomMenu themeDyn mIndexZettel mEditUrl = do
       ffor x $ \case
         Nothing -> blank
         Just indexZettel -> do
-          neuronDynRouteLink (Some . Route_Zettel . Z.zettelSlug <$> indexZettel) ("class" =: "item" <> "title" =: "Home") $
+          neuronDynRouteLink (Some . Route_Zettel . Z.zettelSlug <$> indexZettel) ("class" =: "item" <> "title" =: "Home") $ do
             semanticIcon "home"
+            text " Home"
```
srid, line 156: Why do we need a space before "Home"? Can you post a screenshot of the result?

rvl: Sorry, forgot screenshots. I updated the PR description.

I added the space because it looked unbalanced without it. The reason was that semantic-ui examples have space between the icon and label (otherwise the HTML would be unreadable), but reflex won't add space unless you ask.

But investigating the semantic-ui example further, it looks like there is a css rule `.ui.menu .item>i.icon { margin-right: 0.36em; }`. When I inspect the neuron output, there is no such css rule being applied.

Should I change it so that the icon has a right margin, instead of a space?

srid: It'd be best to figure out why semantic UI css is not being applied. Is our HTML incorrect, or do we have some other CSS that is overriding it (and if so can that be fixed/removed)?

Also, the #342 is for putting a label in z-index (now called "Impulse"), rather than the home icon (which I don't think we need?).

rvl: Oh right! Well I wanted to fix it in the Impulse frontend too, but I assumed that the issue was about the bottom buttons, because they are semantic-ui menu items, but it's just a plain bullet list on the Impulse page.

srid:
Closing per https://neuron.zettel.page/next

PR #556, opened 2021-02-14
`neuron search`: highlight matching line

jilleJr:
I found searching quite difficult, and had this in mind for a while.

These are only bash changes on how neuron interacts with `fzf` and `bat`

## Changes

- Highlight line with `bat --highlight-line`
- Changed FZF to focus on the line with `fzf --preview-window ':+5-5'`
- Made the focuse'd line configurable via `NEURON_SEARCH_CONTEXT_ABOVE` environment variable, but defaults to `10` (lines above search result) *(edit: changed from 5 to 10)*

## Preview

Before:

[asciicast](https://asciinema.org/a/391208)

After:

[asciicast](https://asciinema.org/a/391205)

### DIFF
--- a/neuron/src-bash/neuron-search
+++ b/neuron/src-bash/neuron-search
@@ -9,6 +9,8 @@ EXTENSIONS=${4}
 OPENCMD=$(envsubst -no-unset -no-empty <<<${5})
 EXACT=""
 
+: ${NEURON_SEARCH_CONTEXT_ABOVE:=5}
+
 # Use --exact for full text search
 if [ -z "${FILTERBY}" ]; then
   EXACT="--exact"
@@ -25,5 +27,6 @@ cd ${NOTESDIR}
   | ( [ ${FILTERBY} ] && LC_ALL=C sort -t: -k1,1 -u || cat )\
   | fzf ${EXACT} --tac --no-sort -d ':' -n ${SEARCHFROMFIELD}.. \
     --preview 'bat --style=plain --color=always --highlight-line={2} {1}' \
+    --preview-window ':+{2}-'"$NEURON_SEARCH_CONTEXT_ABOVE" \
   | awk -F: "{printf \"${NOTESDIR}/%s\", \$1}" \
   | xargs -I % ${OPENCMD} %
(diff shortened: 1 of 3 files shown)


neuron/src-bash/neuron-search
```diff
@@ -9,6 +9,8 @@ EXTENSIONS=${4}
 OPENCMD=$(envsubst -no-unset -no-empty <<<${5})
 EXACT=""
 
+: ${NEURON_SEARCH_CONTEXT_ABOVE:=5}
```
jilleJr, line 12: Where can this be documented? @srid 

Something like

> Changed the context in search view by setting the environment variable `NEURON_SEARCH_CONTEXT_ABOVE`, ex:
>
> ```sh
> # inside my ~/.bashrc
>
> # now search term will be on first line in preview
> export NEURON_SEARCH_CONTEXT_ABOVE=0
> ```

jilleJr, line 12: @srid  Does this need to be able to be disabled btw, back to original "only show the start of the zettel"?

srid:
Great!

1. Does this affect `neuron search` (i.e., without the "-a" argument) in anyway?
2. Could you update CHANGELOG.md and increment the version in neuron.cabal (to `1.9.10.0`)?

neuron/src-bash/neuron-search
srid: Why should this parameter to be made configurable in the first place? Does this affect the behaviour of (non-full-text) `neuron search`?

jilleJr: This only affects the visual behaviour. And yes, it affects both fulltext and exact searching

jilleJr:
> Great!
> 
> 1. Does this affect `neuron search` (i.e., without the "-a" argument) in anyway?
> 2. Could you update CHANGELOG.md and increment the version in neuron.cabal (to `1.9.10.0`)?

1. Yes. It affects it the same way.
2. Sure can!

jilleJr:
Here's a preview of how regular `neuron search` looks:

[asciicast](https://asciinema.org/a/ttd4xTLBZc2lDyqDdfJLAwIfH)

Issue #557, opened 2021-02-14
Support org-roam format


JonathanReeve:
So I know org support has been removed in v2, but it seems to me that there are a lot of great reasons for supporting org-roam file formats in Neuron: 

 - Org-roam has a large user base already, and so Neuron would get a lot of users, too, if it supported org-roam format. 
 - Org-roam doesn't yet have a killer HTML exporter, in my opinion. There are hacky versions out there: @jethrokuan's [brain dump](https://braindump.jethro.dev/), and @rasendubi's [org-braindump](https://github.com/rasendubi/uniorg/tree/master/examples/org-braindump), but @jethrokuan's approach converts all org files to Markdown in order to serve them with Hugo, which seems a bit excessive, and @rasendubi's approach is a brand new org-mode parser in Typescript. 
 - Neuron v1 (and early v2?) is already really close to supporting org-roam files, already, since Pandoc has a built-in Org parser, and there's [an Org reader here](https://github.com/srid/neuron/blob/v1/neuron/src/app/Neuron/Reader/Org.hs), even though it's not very complete. 
 - I just really like the Haskell / Neuron and Rib / Reflex approach to static site publishing. 

So [I started adding some stub functions here on my fork](https://github.com/JonathanReeve/neuron/blob/v1/neuron/src/app/Neuron/Reader/Org.hs). Any help there, either from org-roam contributors or neuron contributors, would be much appreciated, since I'm still fairly new to Haskell.

srid:
To add back org-mode support in neuron, I'd recommend checking out the Plugin system in `master` branch. Support for other text formats is best added as a plugin. Part of this will involve figuring out how to best abstract out (without compromising neuron core's simplicity) the various places where Markdown-assumption is used (eg: when checking for `.md` file extension).

Plugins are defined here: https://github.com/srid/neuron/tree/master/neuron/src/lib/Neuron/Plugin ... a plugin is just a set of hooks, telling neuron to do custom things at various stage of the build process.

However, the Plugin type itself is hardcoded to assume markdown format (see for example the `_plugin_markdownSpec` hook). So we have to figure out a way to treat source formats polymorphically in the plugin type.

srid:
Okay, I did some WIP prototype, and realized that org-mode support can be done as builtin -- doesn't have to happen as a plugin ... however we have to think in terms of how it impacts existing plugins. 

[...]

PR #559, opened 2021-02-21
WIP v2 org-mode support

srid:
Extremely rough but working org-mode support for demo and testing. #557

### DIFF
--- a/neuron/neuron.cabal
+++ b/neuron/neuron.cabal
@@ -1,7 +1,7 @@
 cabal-version: 2.4
 name: neuron
 -- This version must be in sync with what's in Default.dhall
-version: 1.9.10.0
+version: 1.9.11.0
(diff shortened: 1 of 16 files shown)


srid:
Links don't work currently. In v1, I believe we allowed links like `[[z:/<id>]]`. This requires some new consideration, given v2 changes. Org-mode users are encouraged to come up with a final solution. 

Whatever syntax we decide on, it should be transformed in the AST inside of `parseOrg` using `Text.Pandoc.Walk` (to be consistent with what the markdown reader returns), so that the rest of neuron (esp. Links and Tags plugins) work normally without any change. 

/cc @felko @JonathanReeve @alex-a-soto

Jonathan - if you are interested in hacking on this PR, feel free to take over. No guarantee when / if I'll get back to this.

PR #561, opened 2021-02-23
Use colors in logs only if terminal supports it.

### LINKED ISSUE #555: ANSI colors in emacs look ugly
When [neuron-mode](https://github.com/felko/neuron-mode) uses neuron executable for generating HTML, its output in the `*compilation*` buffer look like:

    [90m   Plugins enabled: links, tags, uptree[0m
    [90m   Loading directory tree (84 .md files) ...[0m
    [90m   Building graph (84 notes) ...[0m
    [90m   Building route data ...[0m
    [90m   Rendering routes (85 slugs) ...[0m
    [94m💾 impulse.html[0m
    [94m💾 forth.html[0m
    [94m🥅 2 files updated in /home/ark/zettelkasten/.neuron/output[0m

If I understand right, [System.Console.ANSI](https://hackage.haskell.org/package/ansi-terminal-0.10.3/docs/System-Console-ANSI.html) has function `hSupportsANSI`, which can determine this situation. Maybe it would be better not to use colors if this function returns False.

forketyfork:
Closes #555.

### DIFF
--- a/neuron/src/app/Neuron/CLI/App.hs
+++ b/neuron/src/app/Neuron/CLI/App.hs
@@ -30,6 +30,7 @@ import qualified Neuron.LSP as LSP
 import qualified Neuron.Version as Version
 import Options.Applicative
 import Relude
+import System.Console.ANSI (hSupportsANSI)
 import System.Directory (getCurrentDirectory)
 
 run :: (Bool -> App ()) -> IO ()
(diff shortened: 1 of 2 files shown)


srid:
Looks good, @forketyfork - but instead of changing the semantics of the `color` function itself (which now behaves unexpectedly compared to how it is named), you'll want to just call it conditionally:

```haskell
-- Instead of: color c
(bool id color useColors) c
```

(We might as well rename `color` to `colorize` to avoid confusion)

forketyfork:
@srid thanks, this makes sense, I've updated the code.

srid (approved):

Issue #568, opened 2021-02-24
Client-side full-text search


srid:
Depends on #567

Checkout how mdBook does full-text search: https://rust-lang.github.io/mdBook/ ... it seems to be using http://elasticlunr.com/docs/index.html

flyinggrizzly:
I can maybe look at this--I'm not a search expert, but the ElasticLunr docs look pretty friendly.

The issue says this depends on the seemingly more basic reinstatement of client-side search, and I'm understanding the idea here is to use the `cache.json` file to build up an ElasticLunr index, and then use it for search. Is that about right?

I'm not a Haskell developer, so I won't be much use in the first step, but once there is a cache file, I could pick this up to start it rolling.

And if so, my thought would be to:

- spike out the index using vanilla JS, in the browser
- spike out the searcher using vanilla JS

Then, once the general approach is validated

- look into TypeScript (one of my primary languages), or PureScript (unfamiliar, but I imagine I could pick it up) for the searcher
- look at extending and tuning searcher functionality

And, finally, if generating the search index in the client is too slow, investigate what moving that to a server-side (pre)-render might look like.

Given that this depends on #567, please feel no pressure from me. I'll start looking at mdBook and ElasticLunr in the meantime--they look quite interesting.

srid:
@flyinggrizzly Feel free to to prototype a working version. I haven't done any research on this, and I don't know how the two issues are exactly related, but I imagine we can do even the non-fulltext metadata search (#567) using elasticlunr itself?

/cc @TheMatten who expressed some interest in the PureScript part.

srid:
What does the elasticlunr index look like? If it is not too complex, I can probably quick do the index generation part done in Haskell. Otherwise, I imagine you'd fiddle with `cache.json` ...

srid:
By the way, cache.json is already available in the static site, eg: https://neuron.zettel.page/cache.json

flyinggrizzly:
Oh weird. I must have just had a generation locally where `cache.json` didn't show up since I was getting 404s fetching it. But yea, if `cache.json` is available, I can use that to generate an index at page load (I think).

The index document looks [structurally pretty straightforward](https://rust-lang.github.io/mdBook/searchindex.json), but the `index` element itself isn't exactly simple. I haven't looked at the ElasticLunr index generation code, so I don't know how it would look to reimplement it in Haskell.

[...]

Issue #567, opened 2021-02-24
Bring back client-side search


srid:
Originally we had a JavaScript based search that used `cache.json` to search zettels by title or tag. It got replaced by GHCJS-based Impulse search, but that turned out to be *very slow* larger notebooks, so it got removed. So let's reinstate the old search feature, based on JavaScript or any other language that compiles directly to JS (without runtime overhead), such as TypeScript and PureScript.

srid:
https://blog.fission.codes/stork-search-james-little/

> Stork is two things. First, it's an indexer: it indexes your loosely-structured content and creates a file that you can upload to your web server. Second, it's a Javascript + WebAssembly frontend for that index file: Stork will hook into an <input> on your web page, download the index you've specified, and display the best search results immediately to your user, as they type. The precomputed index and WebAssembly frontend module make the entire Stork engine very good, and very fast.

---

Some inspiration for the UX part of search: https://vitepress.vuejs.org/ (check the top right corner)

Issue #358, opened 2020-08-24
Support for "or" and "not" in tag queries


Remich:
Hi,

I figured by trial and error (_missing in documentation_) that automatic links support the "and-operator": e.g.: `[[z:zettels?tag=science&tag=math&timeline]]`

It would also be nice to have support for "or" and "not" operators, and maybe parenthesis, in order to create arbitrarily complex search queries for auto-linking.

This would allow for a better auto-linking experience.

srid:
We could add a `q` parameter to support complex queries, for example:

```
[[z:zettels?q="(tag:science OR tag:physics) -(tag:math)"]]
```

Which in flexible links of #312 would look like:

```
{{zettels | q = (tag:science OR tag:physics) -(tag:math) }}
```

srid:
FTR, `and` / `or` differentiation is available in the code via the `TagQuery` type as of #500. That type will have to be enriched further in order to support this feature (at this point, however, the whole query thing must refactored into a plugin of its own).

srid:
This should get implemented after #407

s9gf4ult:
I would 

```
{{zettels | q = (tag.science | tag.physics) & !(tag.math) }}
```

* `&` for AND
* `|` for OR 
* `!` for NOT
* `()` for brackets

Also, I would allow to query by arbitrary zettel's metadata. 

Consider this meta

``` yaml
issue:
  status: "bug"
```

You could query it by

```
{{ zettels | q = issue.status = "bug" }}
```

So, there must be also

* `"` for string literals
* `true` `false` for boolean literals
* `=` for comparison of two metadata values

srid:
Don't see the value of favouring C-like operators instead of using standard search query operators used by other apps and sites, but doing the below is a good idea.

>  I would allow to query by arbitrary zettel's metadata

srid:
Closing per https://neuron.zettel.page/next ... to further discuss this idea, post in https://github.com/srid/emanote/discussions/48

PR #575, opened 2021-02-28
Spike elasticlunr in docs site

flyinggrizzly:
This PR is a spike at adding client-side search using
[ElasticLunr](https://elasticlunr.com).

This PR message, and the code, are still WIP. I'll keep this up to date as I go.

At the moment it is still an experiment, manually adding in scripts in the
`doc/` site.

Because `cache.json` does not have the note bodies/content, we can't build in
full-text search yet.

I'm not sure if that's a problem at this point, since it may make sense to move
the search index generation out of the browser and into Neuron itself, so that
an index can be requested by the browser instead of being constructed on the
fly.

ElasticLunr appears to have some support for [jsonifying its
indices](http://elasticlunr.com/docs/index.js.html#toJSON), though I
haven't looked into the process of hydrating these yet. [It looks pretty
straightforward though](http://elasticlunr.com/docs/index.js.html#load).

## To do:

- [x] display results in UI instead of logging to console
- [ ] test jsonifying/rehydrating an index to see how it would behave if we were
    to build the index server-side
- [x] investigate "pipelines" (stop-word removal, stemming)
- [x] investigate plugins (is this the same as above?)
- ~~determine good/optimum search configs (and suppress default config warning)~~
- [x] spike "advanced search" functionality, possibly with `#fieldSearch`
  - may want its own `search.html` page, so that might want to be part of
      displaying results before moving too far ahead
- [ ] Handle pretty-url configuration in search
- [ ] Add date (and other field?) search
- [ ] Enable advanced search options (tags, dates, etc) based on relevant plugin settings in Neuron

Once those checks are done, I think things'll be in a position to make a better
decision about whether this makes sense for Neuron.

References #568

For #567

---

### Next immediate steps

- [x] use saved cache from https://lost-frequencies.eu/impulse.html to make sure we're not hitting performance issues
- [x] remove 3-char limit in search function
- [x] blindly copy elasticlunr settings from mdBook and test
  - [x] investigate each one and remove any unnecessary options
  - [x] if this doesn't fix the `insta` search issue, keep investigating
- [x] replace Rails-style syntax with Obisidian-style syntax in URL parsing
- [ ] move search box out to footer/header and remove `seach.md`

### DIFF
--- a/doc/head.html
+++ b/doc/head.html
@@ -1,0 +1,1 @@
+<script src="/static/elasticlunr.js"></script>
--- a/doc/head.html
+++ b/doc/head.html
@@ -1,1 +1,2 @@
 <script src="/static/elasticlunr.js"></script>
+<script src="/static/search.js"></script>
--- a/doc/search.md
+++ b/doc/search.md
@@ -1,6 +1,24 @@
 
 ```{=html}
 <div id="search-box">
+  <style type="text/css">
+    li.search-result-entry {
+      margin-bottom: 25px !important;
+      list-style: none;
+
+      padding-left: 10px !important;
+      border-left: 3px solid lightblue;
+    }
+
+    div.search-result-entry--tags {
+      margin-top: 5px;
+    }
+
+    div.search-result-entry--tags h5 {
+      margin-top: 5px !important;
+      margin-bottom: 0px !important;
+    }
+  </style>
   <input id="search-input" class="ui input" placeholder="Type to search" />
   <ul id="search-results">
   </ul>
--- a/doc/search.md
+++ b/doc/search.md
@@ -1,0 +1,8 @@
+
+```{=html}
+<div id="search-box">
+  <input id="search-input" class="ui input" placeholder="Type to search" />
+  <ul id="search-results">
+  </ul>
+</div>
+```
--- a/doc/search.md
+++ b/doc/search.md
@@ -1,3 +1,0 @@
----
-unlisted: true
----
--- a/doc/search.md
+++ b/doc/search.md
@@ -1,26 +1,3 @@
-
-```{=html}
-<div id="search-box">
-  <style type="text/css">
-    li.search-result-entry {
-      margin-bottom: 25px !important;
-      list-style: none;
-
-      padding-left: 10px !important;
-      border-left: 3px solid lightblue;
-    }
-
-    div.search-result-entry--tags {
-      margin-top: 5px;
-    }
-
-    div.search-result-entry--tags h5 {
-      margin-top: 5px !important;
-      margin-bottom: 0px !important;
-    }
-  </style>
-  <input id="search-input" class="ui input" placeholder="Type to search" />
-  <ul id="search-results">
-  </ul>
-</div>
-```
+---
+unlisted: true
+---
--- a/doc/static/elasticlunr.js
+++ b/doc/static/elasticlunr.js
@@ -1,0 +1,2507 @@
+/**
+ * elasticlunr - http://weixsong.github.io
+ * Lightweight full-text search engine in Javascript for browser search and offline search. - 0.9.5
+ *
+ * Copyright (C) 2017 Oliver Nightingale
+ * Copyright (C) 2017 Wei Song
+ * MIT Licensed
+ * @license
+ */
+
+(function(){
+
+/*!
+ * elasticlunr.js
+ * Copyright (C) 2017 Oliver Nightingale
+ * Copyright (C) 2017 Wei Song
+ */
+
+/**
+ * Convenience function for instantiating a new elasticlunr index and configuring it
+ * with the default pipeline functions and the passed config function.
+ *
+ * When using this convenience function a new index will be created with the
+ * following functions already in the pipeline:
+ * 
+ * 1. elasticlunr.trimmer - trim non-word character
+ * 2. elasticlunr.StopWordFilter - filters out any stop words before they enter the
+ * index
+ * 3. elasticlunr.stemmer - stems the tokens before entering the index.
+ *
+ *
+ * Example:
+ *
+ *     var idx = elasticlunr(function () {
+ *       this.addField('id');
+ *       this.addField('title');
+ *       this.addField('body');
+ *       
+ *       //this.setRef('id'); // default ref is 'id'
+ *
+ *       this.pipeline.add(function () {
+ *         // some custom pipeline function
+ *       });
+ *     });
+ * 
+ *    idx.addDoc({
+ *      id: 1, 
+ *      title: 'Oracle released database 12g',
+ *      body: 'Yestaday, Oracle has released their latest database, named 12g, more robust. this product will increase Oracle profit.'
+ *    });
+ * 
+ *    idx.addDoc({
+ *      id: 2, 
+ *      title: 'Oracle released annual profit report',
+ *      body: 'Yestaday, Oracle has released their annual profit report of 2015, total profit is 12.5 Billion.'
+ *    });
+ * 
+ *    # simple search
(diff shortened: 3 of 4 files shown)


srid:
> Because cache.json does not have the note bodies/content, we can't build in
full-text search yet.

At least we can prototype a working version of metadata search (i.e., search by title & tags), right?

flyinggrizzly:
> At least we can prototype a working version of metadata search (i.e., search by title & tags), right?

That's what I was thinking, yea.

flyinggrizzly:
@srid I've just pushed up some commits with a basic UI (screenshot added to PR description). This may be all I have time for this weekend, but my plan is to hit the plugins/pipeline next.

flyinggrizzly:
Just pushed up a few more commits that add handling of URL params for search. This is a nice-to-have, but also saves me the trouble of adding in a UI for advanced search before I fully understand it.

Currently, I'm using this to attempt searching only the `tags` field, with `/search.html?search[tags]=["root"]`.

I've got the simple general search working, with `/search.html?search=Install`, which returns the same results as if you navigated in and entered "Install" into the input, which is 👍.

Field-scoped search is going to take some more research on my part though.

srid:
Cheers, I'll take a peek.

We should test the search perf on large Zettelkastens like https://lost-frequencies.eu/impulse.html 

@zettelzottel Is the source of your public zettelkasten available somewhere?

zettelzottel:
@srid no the source is not public. it is hosted on private gitea instance

flyinggrizzly:
@zettelzottel if you're OK with me signing up I can start testing with it, but if you'd rather not have some rando looking at the source I totally understand.

srid:
Maybe use https://lost-frequencies.eu/cache.json in `./doc`'s zk just for testing ...

srid:
@flyinggrizzly Search fails for me with this warning in the console

image

Anything in particular I should do when running neuron on ./doc? I access the search HTML using: http://localhost:8080/search.html

srid:
It works (as in it displays the results) when I go to a URL with query like http://localhost:8080/search.html?search=Install (the console warning above still appears), but not when I type a query and press Enter (nothing happens then).

flyinggrizzly:
Yea, that search config warning is still something to look at--I'll add it to the to-do list. Related, I've added in a very basic and probably problematic config for the fieldSearch call that also needs to be actually looked at--so far I've just added enough to get a successful call of the method.

srid:
"instal" works, but "insta" (or "inst") returns nothing. I'd expect search to work even if you type only 3 characters of a word. I suppose this is a matter of configuring elasticlunr? And yea search works without having to hit enter. 

I think it would be good to see tag-based search; we can then create permalinks to search page that lists all zettels tagged with a tag (and link to it from tag links). Something like `/search.html?q=tag:foo`

Also, what kind of index file can neuron generate that would help the frontend JS maximally (in regards to perf)? Assuming metadata only.

flyinggrizzly:
Just checking in--this weekend ended up being overwhelming so I didn't get the time I planned to work on this. But, before it goes too long, here's my thoughts/plans:

## Search by tag

Tag search is already configured in the URL params handling, using `search[tags]=[tag1,tag2]`, which should allow linking by tag, and also linking to custom searches of multiple tags. I can't remember if elasticlunr defaults to AND or OR conjunctions, but it can be set either way.

I went with a Rails-esqu format for the params, but can very easily switch to `?tag:[foo]`/`?tag=[foo]`/`?tag:foo` etc.

At the moment, only 1 tag is ever searched at a time, but I thought that building in the option of multiple would help in the future, and doesn't really cost anything now.

Do you have a query-string syntax preference? At the moment, cases we need to handle are:

- general (keyword) search. Currently `?search=search%20terms`
- tag-search. Currently `?search[tags]=[tag1]`

Also, based on the current config, we could also have title search, slug search, or date search (which is a bigger can of worms, but I suspect elasticlunr might have some decent date logic if it's based on Lucene syntax. But not sure). Ideally these would have similar syntax to tags, at least in the naming of the query parameter (so `search[title]` or `search[date]`). I wasn't planning on working on these for now, since tags and general search seem the priorities, but worth considering when we look at how we want the query string formed.

NOTE: while tag search is set up in the URL params handling, it's not working--I still need to get my head around `fieldSearch`

## Minimum search string length

That does look like an elasticlunr thing--mdBook search fires on any number of keystrokes. 

I had also added a minimum search string length of 3 which I'll remove too, just in case that's contributing somehow.

## Server generated support

For now, I still think it'd be best to keep everything in the browser--the cache is supplying everything we need still.

At some point the search index may need to have behavior configured based on plugins that are active, but those are already listed in the cache so that's still OK.

srid:
- For search query format, see https://github.com/srid/neuron/issues/358#issuecomment-679351864 .. so perhaps using a generic `?q="foo tag:someTag" (which allows complex queries like like `?q="foo title:Foo (tag:tag1 OR tag:tag2)"` would be good for long-term. To begin with, we don't necessarily have to support from the get-go URL queries other than a singular tag filtering (for use in tag links), i.e., `?q=tag:someTag`.
- Re: server generated cache ... okay if cache.json is good enough, let's use that. We can optimize index performance (esp. for larger notebooks) latter anyway, once the general search implementation is in place.

---

I see that you have a `search.md` zettel that contains the HTML. Ultimately, neuron could generate `search.html` with this HTML/JS in it, but a better approach is to put the search box in the footer action bar in all zettel pages, and do the search right there, and display the results inline ... just like mdbook does currently. Then we don't need a dedicated search page. I imagine I would do this in PureScript (or TypeScript or whatever), using your `search.html` and `search.js` as reference implementation (which this PR would provide as a proof of concept).

srid:
>  I imagine I (or somebody) would do this in PureScript (or TypeScript or whatever)

Actually you know what. Let's just do this in plain JavaScript, to begin with. I could modify neuron to use the current `search.js` in all generated pages.

flyinggrizzly:
That makes sense to me. Do you have a minimum browser target in mind? I've been working under the assumption that some compilation might be used so I've been using plenty of ES6--it wouldn't be a problem to manually roll it down to ES5 compatible syntax etc, but I'll just need to check out what works.

Also, that query syntax works--I'll update it to use that.

For `search.md`, this was just a fast way of getting up and running--once search behavior is working then I was thinking I'd move it to being present on every page and dropping down over page content. We'll probably want a dedicated (static) search page for linking out to tag queries etc, but that's also pretty easy.

And for the cache stuff, my guess is that it'll be faster/more performant to have a server generated search index as mdBook does, but I don't think the search function is stable enough (or even known to be actually work for Neuron 😅) to warrant putting more of your time into it yet.

srid:
Some inspiration for the UX part of search: https://vitepress.vuejs.org/ (check the top right corner)

flyinggrizzly:
## "ins" as search term

This wasn't working because of a search settings issue--the `searchOptions` parameter for index searching takes an optional `expand` attribute, which when set to `true` causes the search terms to be expanded. The example is that `"micro"` expands to "microscope" and "microwave", which is the behavior that we want.

## Pipeline

On closer inspection, `mdBook` is using the default pipeline settings, and so are we--I'm making no changes until we determine we need to.

## Tag-search

Fixing the term-expansion also made me realize that perhaps `#fieldSearch` is not the tool we want for tag-search--we might be better off with a general search and using the field `boost` attributes to suppress everything except the tags:

```javascript
const searchOptions = {
  bool: "AND",
  expand: true,
  fields: {
    title: { boost: 0 }, // set to 1 on general search
    tags: { boost: 1 }, // also set to 1 on general search, but now the only field with a boost > 0
  }
}
```

I've still got some time in me today for this so I'll keep going (and editing this comment), but wanted to document progress today so far.

PR #585, opened 2021-03-19
Add devonmorris.dev

DevonMorris:
Really enjoying this tool so far and learning more about the zettelkasten method. I'd like to share my site as an example.


srid:
Looks great!

PR #586, opened 2021-03-20
[WIP]: Add Windows Support

levirs565:
Add Windows support using Haskell Stack.


srid:
`neuron search` uses `executeFile` and then invokes a bash script ([neuron-search](https://github.com/srid/neuron/blob/master/neuron-search)). We can maybe rewrite this bash script in pure Haskell (while enumerating zettels using neuron's graph), but the Windows machine will still need `fzf` and `bat` installed (Windows binaries for which are available).

levirs565:
I made neuron-search script can run in Busybox's bash (aka ash). `neuron search` is work fine. This is current state:

- `neuron new`, work.
- `neuron search`, work. With Busybox's Bash, envsubst-go, less, rg and fzf.
- `neuron query`, work.
- `neuron gen`, not work.
- `neuron gen -wS`, not work. When server is open in browser, browser only loading web.
- `neuron open`, cannot test. `neuron gen` must work before this tested.

*NOTE*: Those tested when all plugin is disabled.

srid:
The use of `callProcess` in place of `executeFile` may break things on Linux. You may want to test `neuron new -e` and `neuron search -e` (with EDITOR environment set to a command line text editor like vim) on Linux or WSL. 

The invalid relPath error can be fixed by making `tagFromPath` in DirTree.hs portable.

Not sure why `neuron gen` would hang after printing "Loading directory tree" ...

srid:
I think for native Windows, it is mainly important for the `neuron gen` command to work well.

It is okay if `neuron search` and `neuron new` remain broken on native Windows, and we can document this as a (temporary?) limitation. 

I really hope to have native Windows support a part of the 2.0 release.

levirs565:
I gave up adding Windows Support for neuron. However, I create another note talking manager named **kasten** written in TypeScript. You can see kasten in https://github.com/levirs565/kasten-note

Issue #485, opened 2020-11-29
Provide a flake.nix


lboklin:
Offer a Nix flake so that neuron can be added to flake-based system configurations. `nix flake`s is an experimental feature of Nix so it would have to be optional, but it's a great feature that will become the standard method for revision control in nix stable eventually.

I was unsuccessful in adding neuron to my flake-based system - it appears the niv pinning is incompatible with flakes.

https://nixos.wiki/wiki/Flakes

Edit:
I managed to add neuron by adding a nixpkgs revision to my inputs which has a working neuron in haskellPackages and put it in an overlay:
```nix
{
  inputs = {
    # ...
    nixpkgs-neuron.url = "github:NixOS/nixpkgs/c83e13315caadef275a5d074243c67503c558b3b";
  };

  outputs = inputs: with inputs; {
    nixosConfigurations.my-machine = nixpkgs.lib.nixosSystem {
      # ...
      modules = [
        ({...}: {
          nixpkgs.overlays = [
            (final: prev: {
              # ...
              inherit (nixpkgs-neuron.legacyPackages.x86_64-linux.haskellPackages) neuron;
            })
            # ...
          ];
      })
      # ...
      ];
    };
  };
}
```

After that I just used it as a top-level package in my home-manager (as a NixOS module) config.

srid:
I'm not famliar with flakes. But would you like to contribute a PR for adding to documentation? We can add it under https://neuron.zettel.page/2012401.html

DerGuteMoritz:
FWIW, since the version in `nixpkgs` is a bit old, I wanted to use a more recent one from upstream. Here's how I was able to make it work with the current `master` at the time of writing:

```nix
{
  inputs.neuron-notes-src = rec {
    url = "https://github.com/srid/neuron/archive/2547c2132783c6909b6f15837ccd92988c389e8e.tar.gz";
    flake = false;
  };
  # ...

  outputs = { neuron-notes-src, ... }: {
    nixosConfigurations.my-machine = nixpkgs.lib.nixosSystem {
      # ...
      modules = [
        ({...}: {
          nixpkgs.overlays = [
            (final: prev: {
              neuron-notes = (prev.callPackage "${neuron-notes-src}/project.nix" { }).neuron;
            })
            # ...
          ];
        })
        # ...
      ];
    };
  };
}
```

Using version `733e537a8ad76fd355b6f501127f7d0eb8861775` of `nixpkgs`.

srid:
[...]

PR #594, opened 2021-04-11
flake.nix: add basic support

TheKK:
This should solve #485.

With this `flake.nix`, we can do...

- use `nix build` to build
- use `nix shell` to enter build shell
- use `nix run` to run `neuron` even without cloning project
  - for instance, `nix run "github:TheKK/neuron#neuron" -- open`

I also add `result-data` to `.gitignore` since it was created after running `nix build`. This make you to run `nix-build` to build same derivation after `nix build`.

If there's anything I can improve, please let me know!

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,9 +4,13 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/fcab19deb78fbb5ea24e19b133cf34358176396a";
     flake-utils.url = "github:numtide/flake-utils";
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
   };
 
-  outputs = { self, nixpkgs, flake-utils }:
+  outputs = { self, nixpkgs, flake-utils, ... }:
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs { inherit system; };
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,24 @@
+{
+  description = "Future-proof note-taking and publishing based on Zettelkasten";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/fcab19deb78fbb5ea24e19b133cf34358176396a";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+        project = import ./project.nix { inherit pkgs; };
+
+      in rec {
+        packages = { neuron = project.neuron; };
+        defaultPackage = packages.neuron;
+
+        apps = { neuron = flake-utils.lib.mkApp { drv = packages.neuron; }; };
+        defaultApp = apps.neuron;
+
+        devShell = project.shell;
+      });
+}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,10 @@ jobs:
       DOCKERTAG: latest
     steps:
       - uses: actions/checkout@v2
-      - uses: cachix/install-nix-action@v12
+      - uses: cachix/install-nix-action@v13
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
       # This also runs nix-build.
       - uses: cachix/cachix-action@v8
         with:
@@ -25,7 +28,13 @@ jobs:
       - name: Build 🔧
         run: |
           nix-build -j4 --no-out-link ci.nix
+          # Test default.nix
           nix-build -j4
+          # Test flake.nix
+          nix build --out-link flake-result
+      - name: Verify output of default.nix & flake.nix
+        run: |
+          [ $(readlink result) = $(readlink flake-result) ]
       - name: Retrieve neuron version
         run: |
           echo "NEURONVER=$(./result/bin/neuron --version)" >> $GITHUB_ENV
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dist-*
 .ghc.environment.*
 result
 result-2
+result-data
 .shake
 .neuron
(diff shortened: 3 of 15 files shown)


TheKK:
Do you think that I should modify github workflow in this PR? Just notice that.

flake.nix
```diff
@@ -0,0 +1,32 @@
+{
+  description = "Future-proof note-taking and publishing based on Zettelkasten";
+
+  inputs = {
+    # Please sync revision with ./dep/nixpkgs/github.json
+    nixpkgs.url =
+      "github:nixos/nixpkgs/fcab19deb78fbb5ea24e19b133cf34358176396a";
```
srid, line 7: Can `edolstra/flake-compat` be used to avoid having to do this? See https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix

I'd ensure DRY when switching to flakes.

More generally, we can get rid of `./dep/nixpkgs` allowing flakes to pin nixpkgs if that's doable.

flake.nix
```diff
@@ -0,0 +1,32 @@
...
+        neuron = project.neuron;
+        neuronApp = flake-utils.lib.mkApp { drv = neuron; };
+
+        shell = project.shell;
+
+      in {
+        packages = { inherit neuron; };
+        defaultPackage = neuron;
+
+        apps = { hello = neuronApp; };
```
srid, line 27: Why "hello"?

srid:
@TheKK I took a quick look and added some comments, thanks for your PR! Yes, CI should be updated. 

Note that I think we should probably migrate entirely to flakes, and then use `edolstra/flake-compat` to provide a default.nix and shell.nix that uses flake.nix. Although it would be good to do that in this PR, it is fine if that happens outside of it - as "stage 2". At minimum, we should hover normalize the nixpkgs pinning in one place, as it is easy to forget updating it in both locations.

flake.nix
```diff
@@ -0,0 +1,32 @@
...
+      "github:nixos/nixpkgs/fcab19deb78fbb5ea24e19b133cf34358176396a";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+
+        project = import ./project.nix { inherit pkgs; };
```
srid, line 16: I believe this should inherit `system` as well.

flake.nix
TheKK: Sure! That sound good. Let me try the `flake-compat` solution first.

flake.nix
TheKK: Sorry about that, I'll fix it.

flake.nix
TheKK: if i understand correctly, only `default.nix` accepts `system` attribute and its only purpose is to eval `import ./dep/nixpkgs { inherit system; }`, which is same as `import nixpkgs { inherit system; }` in flake.

Is that the case?

srid: Right, n/m then.

TheKK:
Should we create a directory like `dep` and put dependencies from inputs of `flake.nix` into it?

I found that I have to repeat

```nix
    let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
    in fetchTarball {
        url = "https://github.com/nixos/nixpkgs/archive/${lock.nodes.nixpkgs.locked.rev}.tar.gz";
        sha256 = lock.nodes.nixpkgs.locked.narHash;
      };
```

for all

```nix
import ./dep/nixpkgs {...}
```

which appears too many times. Or is there other preferential way to prevent these repetitions?

srid:
You should be able to pass the `inputs` from flake.nix to project.nix, no?

TheKK:
For default.nix and shell.nix, yes we can. But there are other users of `dep/nixpkgs` like ci.nix and static.nix, which can not (easily) derive from flake.nix.

srid:
Don't worry about static.nix; that's already broken.

ci.nix should work, but it only requires `pkg` (for `recurseIntoAttrs`), so you can pass that easily? Don't worry about repetition; if we can get something working first, it can be refactored near the end.

TheKK:
Please let me know if there's anything I can improve :)

flake.nix
```diff
@@ -0,0 +1,34 @@
+{
+  description = "Future-proof note-taking and publishing based on Zettelkasten";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs";
```
srid, line 5: Let's pin the nixpkgs rev explicitly here,

```
nixpkgs.url = "github:nixos/nixpkgs/fcab19deb78fbb5ea24e19b133cf34358176396a";
```

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+
+        project = import ./project.nix { inherit pkgs; };
+
+        neuron = project.neuron;
+        neuronApp = flake-utils.lib.mkApp { drv = neuron; };
+
+        shell = project.shell;
```
srid, line 23: remove `neuron` and `shell` variables.

I think you might be able to get away with removing the `neuronApp` variable as well, and do something like this further below:

```nix
rec {
  apps = { neuron = flake-utils.lib.mkApp { drv = project.neuron; };
  defaultApp = apps.neuron;
}
```

(note the `rec`).

srid (changes requested):
Looks good to me, left some comments. And a couple more changes as well:

- Merge with `master` and up version in neuron.cabal to `1.9.27.1`
- Create `nixpkgs.nix` that gives a `pkgs` based on the pinned nixpkgs in flake.nix, and use it in static.nix, docker.nix, ci.nix, etc.

TheKK:
Thanks for your advice!

I've merged fixes you mentioned in the comment and rebase this PR onto 602c3de9d61ed9f1277bc2fc0ec01ca912f395df. Please take a look.

PR #612, opened 2021-05-06
Introduce JSONL format for query

psiska:
To support querying by tools which expects whole information on one
line. E.g. Neovim telescope plugin where I need to query neuron DB returns result split on multiple lines. Which is nice for human, however impose burden on program interaction.

This PR introduces the --jsonl flag to 'query' sub-command. If this flag is active one entry will be printed at each line. 
Which is super convenient format for command line processing.

Example
```
neuron/doc $ /nix/store/i5lagbbinqf4i5b77liz3qp5spih98nd-neuron-1.9.27.2/bin/neuron query  --id index
   Plugins enabled: dirtree, links, tags, uptree
   Loading directory tree (49 .md files) ...
   Building graph (49 notes) ...
[
    {
        "ID": "index",
        "Meta": {
            "dirtree": {
                "display": false
            },
            "tags": []
        },
        "Path": "./index.md",
        "Slug": "index",
        "Title": "Neuron Zettelkasten"
    }
]
```
After
```
neuron/doc $ /nix/store/i5lagbbinqf4i5b77liz3qp5spih98nd-neuron-1.9.27.2/bin/neuron query --jsonl --id index
   Plugins enabled: dirtree, links, tags, uptree
   Loading directory tree (49 .md files) ...
   Building graph (49 notes) ...
{"Path":"./index.md","Slug":"index","ID":"index","Meta":{"dirtree":{"display":false},"tags":[]},"Title":"Neuron Zettelkasten"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
   - `neuron query`:
     - query CLI interface has been changed (see docs)
     - JSON output is now pretty printed (using `aeson-pretty`)
+    - option to choose pretty printed JSON (default) or JSONL format (#612)
   - `neuron gen`: Add `--pretty-urls` to remove `.html` suffix in generated URLs (#562)
   - Supress colors if output is not a terminal with color support (#561)
 - Performance improvements:
--- a/neuron.cabal
+++ b/neuron.cabal
@@ -1,6 +1,6 @@
 cabal-version: 2.4
 name: neuron
-version: 1.9.29.0
+version: 1.9.30.0
 license: AGPL-3.0-only
 copyright: 2020 Sridhar Ratnakumar
 maintainer: srid@srid.ca
--- a/src/Neuron/CLI/Parser.hs
+++ b/src/Neuron/CLI/Parser.hs
@@ -96,6 +96,7 @@ commandParser defaultNotesDir now = do
             (strOption (long "slug" <> help "Open the zettel HTML page" <> metavar "SLUG"))
     queryCommand = do
       cached <- switch (long "cached" <> help "Use cached zettelkasten graph (faster)")
+      jsonl <- switch (long "jsonl" <> help "output JSONL format (better for tool processing)")
       query <-
         fmap CliQuery_ById (option zettelIDReader (long "id" <> metavar "ID" <> help "Get a zettel by its ID"))
           <|> fmap (const CliQuery_Zettels) (switch $ long "zettels" <> help "Get all zettels")
--- a/src/Neuron/CLI/Query.hs
+++ b/src/Neuron/CLI/Query.hs
@@ -14,6 +14,7 @@ where
 
 import Colog (WithLog)
 import Data.Aeson (ToJSON)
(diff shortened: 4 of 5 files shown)


srid:
Hey, thanks for the PR. This seems like an useful feature in general (even though I'm not familiar with the telescope plugin), however I'm not sure if it is worth baking into neuron itself? Especially as you can achieve this using a tool like `jq`

```
❯ neuron query 2> /dev/null | jq '.[]'  | jq -c | tail -3
{"ID":"ghcid","Meta":{"slug":"ghcid","tags":[]},"Path":"./ghcid.md","Slug":"ghcid","Title":"ghcid"}
{"ID":"index","Meta":{"tags":[]},"Path":"./index.md","Slug":"index","Title":"Sridhar Ratnakumar"}
{"ID":"ka Project","Meta":{"slug":"ka","tags":[]},"Path":"./ka Project.md","Slug":"ka","Title":"ka Project"}
```

psiska:
I can see your point, in keeping neuron as small as possible (correct me if I am wrong, but I have an impression this is a goal).

You are right that I can achieve that with some other external tools like 'jq'. No problem in here.
On the other hand this change is quite small, contained and does not add other dependencies.

The 'jq' path requires pipe-ing output from one command to another which implies either shell or handling stream by myself (telescope is in lua so no shell is involved). Another factor is that it requires to have 'jq' installed.

So my point is that this change make it easier for me and hopefully for other who will wish to integrate the neuron into their pipelines. ( I am thinking about something like pandoc but for information manipulation)

If it helps to get this feature in I can redesign the flag to be more future-proof.

srid:
Okay, then let's consider these:

- We shouldn't use Set in the implementation. The existing behaviour (without --jsonl flag) should not change. Pass lists instead of Set.
- Instead of various small functions under the where clause, let's just inline them in the case block - and refactor only strictly necessary.
- Use `maybeToList` (from relude), instead of writing your own (m2s)
- In the case block, use something like `bool printPrettyJson (printJsonLine . maybeToList) jsonl $ result`
- The where block would now contain only two functions: printPrettyJson and printJsonLine
- Replace `putTextLn . decodeUtf8` in printPrettyJson with `putLBSLn` for consistency

srid:
By the way, could you share your neovim telescope config? I'm curious about that plugin.

psiska:
So I got back to the patch. Hopefully I got your comments rights. Let's see :)

When it comes to my telescope config. There is nothing special atm. Just binding of default pickers to shortcuts. What I am trying to achieve is to integrate neuron into it (even if I am aware of neuron2.vim).

srid:
@psiska Looks perfect! Just one last change: could increment the version (to 1.9.30.0) in neuron.cabal and add an entry (referencing this PR) in CHANGELOG.md?

(Make sure to merge with master if you haven't already)

psiska:
I've added the entry in the Changelog and rebase against upstream/master.

srid (approved):

PR #624, opened 2021-05-23
Add named wikilink processing

### LINKED ISSUE #373: Allow custom link text in wiki links
**Proposed syntax**:

```
[[ foo | a foo note ]]
```

**Open question**:

Does this play well with notable plain-text Zettelkasten apps (Obsidian, Zettlr)? If not, what would be a more suitable syntax?

hi-learn-triz-and-usit:
Let me know if there are any issues.

I didn't add tests because there are currently no tests for parsing the wikilinks anyway.

This is technically a breaking change if anyone has pipes in their wikilinks, but I don't think that's a huge problem.

There's also both styles of `[[url|name]]` and `[[name|url]]` in use; I went with the former because it works better with the defaults in the recommended VSCode configuration.

Resolves #373

### DIFF
--- a/src/Neuron/Plugin/Plugins/Links.hs
+++ b/src/Neuron/Plugin/Plugins/Links.hs
@@ -29,6 +29,7 @@ import qualified Commonmark as CM
 import qualified Commonmark.Inlines as CM
 import Commonmark.TokParsers (noneOfToks, symbol)
 import Control.Monad.Writer (MonadWriter, tell)
+import Data.List.Split (wordsBy)
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import Data.Some (Some (Some))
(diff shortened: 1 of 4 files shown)


hi-learn-triz-and-usit:
>    Could you update the docs to mention this? https://neuron.zettel.page/linking
>    Also please update CHANGELOG.md and the version in neuron.cabal (1.9.x.y -> 1.9.(x+1).0)

Done.

Anything else I missed?

srid:
@hi-learn-triz-and-usit Yes, that's good.

I tested this locally, but it doesn't work? For this patch,

```diff
--- a/doc/index.md
+++ b/doc/index.md
@@ -8,7 +8,7 @@ dirtree:
 
 ## Getting started
 
-* [[Installing]]#
+* [[Installing|Install neuron]]#
 * [[Tutorial]]#
 * [[Guide]]#
 * [[Examples]]#
```

I get:

image

The HTML generated is,

```html
<a href="Installing|Install neuron" title="folge">Installing|Install neuron</a>
```

Note that - if nothing else, this parser should work: https://github.com/srid/ema/blob/d8e14ab63a9b25f44591c38c929b04b957f1546d/src/Ema/Helper/Markdown.hs#L195-L213

srid:
Oops! Nevermind, PEBKAC. I didn't even checkout your branch. :-P

src/Neuron/Plugin/Plugins/Links.hs
```diff
@@ -327,15 +328,20 @@ wikiLinkSpec =
...
-      pure s
-    cmAutoLink :: CM.IsInline a => Connection -> Text -> a
-    cmAutoLink conn url =
-      CM.link url title $ CM.str url
+      let parts = CM.untokenize <$> wordsBy (\x -> CM.tokType x == CM.Symbol '|') toks
+      case parts of
+        [url] -> pure (url, Nothing)
+        [url, name] -> pure (url, Just name)
+        [] -> fail "Empty wiki-link encountered"
+        _ -> fail ("Multiple pipe ('|') characters encountered; here are the parts: " ++ T.unpack (T.intercalate ", " parts))
```
srid, line 341: I'd think sometimes the user might want to use "foo | bar" as the custom text, so in theory we can allow everything to the right of the first `|` as the custom text. But let's merge the PR in its current state since this is an improvement. 

In emanote, as an aside this is how we do it: https://github.com/srid/ema/blob/d8e14ab63a9b25f44591c38c929b04b957f1546d/src/Ema/Helper/Markdown.hs#L206-L211

src/Neuron/Plugin/Plugins/Links.hs
```diff
@@ -327,15 +328,20 @@ wikiLinkSpec =
...
+    cmAutoLink :: CM.IsInline a => Connection -> (Text, Maybe Text) -> a
+    cmAutoLink conn (url, name) =
+      CM.link url title $ CM.str (url `fromMaybe` name)
```
srid, line 344: The operator use of `fromMaybe` here is unorthodox. For future reference, I'd just write `fromMaybe url name`.

srid (approved):

samwalls:
I found that this breaks if the link is inside a markdown pipe table. _However_  it can be fixed by using this syntax: `[[my-page\|my-text]]`.

Example
```md
|test column 1|test column 2|
|-------------|----------------------|
| apple       | [[my-page\|my-text]] |
```

hi-learn-triz-and-usit:
@samwalls I think that what breaks is not the parsing of the wikilink but the parsing of the table, which I believe is Pandoc's responsibility. Would you mind checking if the same thing happens in emanote?

PR #661, opened 2021-09-30
Replace gitignore with nix-filter

ilkecan:
Hi. I am currently writing a home-manager module for `neuron` (I will create a separate PR soon). The problem is, current source filtering is not enough to avoid unnecessary rebuilds. `gitignore` uses blacklist approach to remove version control related files (maybe others) but that doesn't prevent nix rebuilding when a comment is added to a Nix file. The better solution is to use a whitelisting approach and specify the paths needed to build the derivations. [nix-filter](https://github.com/numtide/nix-filter) is a library for that and I have been using it for a while without any issues. With these changes, there won't be any unnecessary rebuilds anymore.


project.nix
```diff
@@ -34,7 +34,17 @@ let
...
+    neuron = nix-filter {
+      root = ./.;
+      name = "neuron";
+      include = [
+        "neuron-search"
+        "neuron.cabal"
+        (inDirectory "exe")
+        (inDirectory "src")
+        (inDirectory "test")
+      ];
```
ilkecan, line 46: While by including only these paths, both `nix build` and `nix-build` finishes without any error, I am not really familiar with the internals of the project. There might be other files required for some reason, so it is best if you double check this part.

flake.lock
```diff
@@ -33,17 +33,17 @@
...
-        "rev": "10bb036486138341e4c760c4c98ddadef56fff7b",
+        "rev": "08ef0f28e3a41424b92ba1d203de64257a9fca6a",
         "type": "github"
       },
       "original": {
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "10bb036486138341e4c760c4c98ddadef56fff7b",
+        "rev": "08ef0f28e3a41424b92ba1d203de64257a9fca6a",
         "type": "github"
```
ilkecan, line 47: Currently on master, the revision specified with flake.nix and locked by flake.nix  are different. So this happened automatically.

srid: I think using `nix develop` instead of `nix-shell` in the `bin/run` script should prevent that from happening again.

srid:
Eventually (not in this PR) we should get rid of nix-thunk entirely and use flakes inputs instead.

srid:
I actually like the whitelisting feature of nix-filter, so let's get this in.

PR #662, opened 2021-09-30
Provide homeManagerModule as a flake output

ilkecan:
Main change is the added home-manager module, which can be used like 
```
    neuron = {
      enable = true;
      notesDirectory = "/home/${name}/zettelkasten";
      prettyUrls = true;
    };
```
See https://github.com/divnix/devos/issues/106#issuecomment-782914232 for pointers about how to add a home-manager module from a flake to `imports`.

I also removed the unnecessary app definitions, since by default `nix run` will try the packages with the same name. App definitions are mainly useful if there are multiple binaries or you want to name them differently.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,9 +21,6 @@
         packages = { neuron = project.neuron; };
         defaultPackage = packages.neuron;
 
-        apps = { neuron = flake-utils.lib.mkApp { drv = packages.neuron; }; };
-        defaultApp = apps.neuron;
-
         devShell = project.shell;
       });
 }
--- a/flake.nix
+++ b/flake.nix
@@ -11,7 +11,9 @@
   };
 
   outputs = { self, nixpkgs, flake-utils, ... }:
-    flake-utils.lib.eachDefaultSystem (system:
+    {
+      homeManagerModule = import ./home-manager-module.nix;
+    } // flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs { inherit system; };
         project = import ./project.nix { inherit pkgs; };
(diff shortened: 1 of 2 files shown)


ilkecan:
Also `nix flake show` currently produces an error like
```
error: a 'aarch64-linux' with features {} is required to build '/nix/store/0a5kf39y7vpnbgcnis3gp3spbdpndvij-cabal2nix-neuron.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```

But I couldn't fix this easily. In general, I think the nix code is unnecessarily complex. You said we nix-thunk should be removed in favour of flake inputs in the previous PR. Does that mean https://github.com/srid/neuron/blob/master/nixpkgs.nix should be removed? I can try to work on these in a separate PR.

srid:
>  You said we nix-thunk should be removed in favour of flake inputs in the previous PR. Does that mean https://github.com/srid/neuron/blob/master/nixpkgs.nix should be removed? I can try to work on these in a separate PR.

Yes, that would be awesome. You can use https://github.com/srid/emanote for inspiration (by the way, emanote is the successor project to neuron).

flake.nix
```diff
@@ -21,9 +23,6 @@
         packages = { neuron = project.neuron; };
         defaultPackage = packages.neuron;
 
-        apps = { neuron = flake-utils.lib.mkApp { drv = packages.neuron; }; };
-        defaultApp = apps.neuron;
-
```
srid, line 26: Hmm, why did these two attributes get removed?

srid: Oh n/m; just saw your comment on it.

ilkecan: Maybe, they should be part of a different PR but the reason for removal is they are unnecessary, since by default `nix run` will try the packages with the same name. App definitions are mainly useful if there are multiple binaries or you want to name them differently.

